PR #9, opened 2013-06-28
Some improvements

rainwoodman:
Trying to read the SDSS spectra files with fitsio and noticed a few weirdness:

The DR10 Spectra's doesn't have Z_PERSON. But in the repr of fits[2]['Z_PERSON'] 
I see subcolumn 'Z'.
The first commit fixes this. Is is because when ColumnSubset.columns is scalar and improperly handled.

The second commit fixes a core dump when I run
fits[2]['CLASS_PERSON']['CLASS'][:]
It is because when the target is a scalar, strides is NULL and PyArray_STRIDE, core dumps. The fact that array is a numpy scalar is probably also worrisome; but with this modification the code gives the correct result 'QSO' without core dumps.

Another commit disables printing to stdout. Because printing to stdout from a library will mess up with regular outputs from the programs using the library. issuing python warnings might be more appropriate.

Another commit disables explicit relative imports. Python by default looks for modules in . first, thus there is no need to explicitly declear '.' in the imports.

### DIFF
--- a/fitsio/__init__.py
+++ b/fitsio/__init__.py
@@ -20,7 +20,7 @@ from .fitslib import READONLY
 from .fitslib import READWRITE
 from .fitslib import cfitsio_version
 
-from .fitslib import BINARY_TBL, ASCII_TBL, IMAGE_HDU
+from fitslib import BINARY_TBL, ASCII_TBL, IMAGE_HDU
 
 
-from . import test
+import test
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -35,6 +35,9 @@
 // not sure where this is defined in numpy...
 #define NUMPY_MAX_DIMS 32
 
+#define ALLOW_NOGIL //PyThreadState *_save; int _evaltmp123_;
+#define NOGIL(x) ({PyThreadState * _save1=PyEval_SaveThread(); int _evaltmp123_ = (x); PyEval_RestoreThread(_save1); _evaltmp123_;})
+
 struct PyFITSObject {
     PyObject_HEAD
     fitsfile* fits;
@@ -323,6 +326,8 @@ void append_string_to_list(PyObject* list, const char* str) {
 static int
 PyFITSObject_init(struct PyFITSObject* self, PyObject *args, PyObject *kwds)
 {
+    ALLOW_NOGIL
+
     char* filename;
     int mode;
     int status=0;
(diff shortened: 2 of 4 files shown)


rainwoodman:
I've rebased to the current master branch, and added GIL release at function calls to cfitsio.
Otherwise fitsio won't be able to take the advantage of threads (eg via multiprocessing.pool.ThreadPool)

esheldon:
In what cases will cfitsio use multiple threads?  Will it use > 1 thread by default now, or only if using multiprocessing?

For example, would cfitsio use more than one thread on a queue system where the user has been assigned a single core, and has not tried to use multiprocessing?

I'll have a look at your other changes and merge them by hand.

rainwoodman:
Threading doesn't occur on cfitsio level; but cfitsio is compatible with threads (see https://heasarc.gsfc.nasa.gov/docs/software/fitsio/c/c_user/node15.html )

Suppose files reside on a 'potentially slow` file-system (eg lustre or gpfs), using multiple threads can allows overlapping computation with IO. 
For example, if I have a list of fits file names, after the patch one can use

```
filename = ['a.fits', 'b.fits', 'c.fits']
import multiprocessing.pool
pool = multiprocessing.pool.ThreadPool()
def process(filename)
    f = fitsio.FITS(filename)
    # process file f
pool.map(filenames)
```

Python cannot take advantage of this unless the GIL is released. This is a well known issue. As an example, most numpy functions releases GIL in the computation core as well. See http://docs.scipy.org/doc/numpy/release.html (More GIL releases) 

Of course multiprocessing can work around the GIL issue too, but it doesn't hurt to release the GIL as much as possible (plus numpy is also doing so as well).

Most of the other patches are very dated and I can squash them. Agree with you on the import.

esheldon:
I think the example you give is best addressed by running separate jobs for each file rather than separate threads for each file.

It comes at the cost of making the code harder to write and understand.  I will have to remember to every call in this NOGIL.

I'm also worried this will have unexpected consequences.  Can you please also add some unit tests (in test.py) and make sure it works in some realistic situations?

rainwoodman:
I agree the example I gave is indeed too simple. A typical useful case where running multiple jobs become cumbersome is

```
filename = ['a.fits', 'b.fits', 'c.fits']
import multiprocessing.pool
pool = multiprocessing.pool.ThreadPool()
def process(filename)
    f = fitsio.FITS(filename)
    # process file f
    # for example select objects above a threshold
    return list_of_objects

objects = numpy.concatenate(pool.map(filenames))
```

I do not agree with you this patch makes the code harder to read, write or understand.
The NOGIL instruction is similar to the "with nogil:" construct in Cython. It clearly states the purpose: functionality that does not involve python runtime shall not hold the GIL. The flow of the original source code is unaffected.

I will start a new PR with the test cases.

rainwoodman:
Initially I implemented this via a few temporary variables and a comma expression. The code becomes standard C after this change:
```
#define ALLOW_NOGIL int _tmp123_; PyThreadState * _save2;
#define NOGIL(x) (........)
```
It is necessary if the compiler doesn't support 'statement expr'. It is a GNU extension; I haven't put in the necessary ifdef for this.

esheldon:
do you have unit tests completed?

rainwoodman:
Not yet. I started a new PR though.

PR #27, opened 2014-10-24
enabled setup to use system cfitsio

beckermr:
I have added an option to use the system cfitsio distribution instead of recompiling. This option is turned off by default, but is a nice build feature to have. You can pass the option `--use-system-fitsio` directly to the `setupy.py` script. You should also be able to invoke it with `pip` via 
```python
pip install fitsio --install-option="--use-system-fitsio"
```
I know this breaks the bundling rule, but it gives users more flexibility when compiling `cfitsio` locally fails using the default `make` invoked in `setup.py`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,9 +7,10 @@ New Featurees
         by SCAMP and return a FITSHDR object
     - reserved header space when creating image and table extensions
         and a header is being written.  This can improve performance
-        substantially on distributed file systems.
+        substantially, especially on distributed file systems.
     - When possible write image data at HDU creation.  This can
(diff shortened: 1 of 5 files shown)


beckermr:
OK. I am going to close this since it appears this will not be used and now my branch is way behind.

esheldon:
I think we should probably make this work if you are still interested.

I think some packages have a magical way to determine that fitsio has been loaded (even if it isn't going to conflict) and will raise exceptions.

beckermr:
I didn't follow your comment, but am happy to reimplement this feature. There is now a merge conflict, so this will need to be redone carefully.

PR #35, opened 2015-04-03
Nogil

rainwoodman:
replaces the old PR.


rainwoodman:
@esheldon I added a simple test case and fixed a few bugs.

fitsio/fitsio_pywrap.c
```diff
@@ -1597,8 +1637,7 @@ int write_string_column(
 
     strdata = malloc(nelem*sizeof(char*));
     if (strdata == NULL) {
-        PyErr_SetString(PyExc_MemoryError, "could not allocate temporary string pointers");
-        *status = 99;
+        *status = 113;
```
rainwoodman: This avoids Python Runtime in function write_string_column.

fitsio/fitsio_pywrap.c
```diff
@@ -1882,14 +1928,18 @@ PyFITSObject_write_columns(struct PyFITSObject* self, PyObject* args, PyObject*
...
+    }
+_fitsio_pywrap_write_columns_bail1:
     free(is_string); is_string=NULL;
     free(colnums); colnums=NULL;
     free(array_ptrs); array_ptrs=NULL;
     free(nperrow); nperrow=NULL;
     free(fits_dtypes); fits_dtypes=NULL;
-    if (status != 0) {
+    if (status > 0) {
+        /* negative is our errors */
```
rainwoodman: -1 means PyErr is already set.

esheldon:
more generally, why are some fits_write* calls wrapped in NOGIL and others are not?

rainwoodman:
1. It is not to avoid multiple threads writing to the same file. It is up to the user to avoid that. there is no cfitsio or fitsio level protection.
2. some fits_write are not in GIL, because 
1) the parent function doesn't depend on Python runtime. In that case, the call to the parent function is wrapped in NOGIL. 
2) the entire region is released from GIL by Py_BEGIN_ALLOW_THREADS. I had one or two cases of these to avoid frequently releasing and grab GIL in loops.
3. The basic functionality seems to be working fine, but I agree with you to give it more time till the branch equilibriums. 

4. I would write a few more test cases, especially cases where cfitsio would raise an exception. Could you recommend a few cases?

rainwoodman:
Random segfaults at testImageWriteRead (fitsio.test.TestThreading) on travis. Need some investigation -- probably a very easy one, e.g. out of memory.

esheldon:
I still don't like the complexity that this new code creates. Every new function must be wrapped by this macro.  I think this makes development harder.

Also I personally do not think it is needed.  For most large scale processing, where performance matters, one would run N jobs rather than N threads for a single job.

rainwoodman:
I agree with you about the 'in practice' part.

I think the main driving force on my side was a matter of pureness: if the GIL can be released then it shall be.  Also one very probable reason for preferring n jobs to n threads is precisely because the GIL is not released -- isn't it?

Since pywrap isn't going to change much I think we can just leave this here as an illustration what has to be done to release the GIL. And incorprate these changes in the future if it ever happens to be rewritten, or if there is sufficient demand on no gil.

esheldon:
I think N jobs is preferred because it is easier to schedule single core jobs than multi-core jobs on  high throughput systems

Of course nersc is a counter example as HPC instead of HTC

beckermr:
I am going to close this PR for now. It is very out of date and given the discussion it is pretty clear that it won't be accepted. Feel free to reopen!

PR #36, opened 2015-04-03
convert print to warnings.

rainwoodman:
avoid interfering with stdout in a non-controlable way.

### DIFF
--- a/fitsio/__init__.py
+++ b/fitsio/__init__.py
@@ -22,5 +22,6 @@ from .fitslib import cfitsio_version
 
 from .fitslib import BINARY_TBL, ASCII_TBL, IMAGE_HDU
 
+from .fitslib import FITSRuntimeWarning
 
 from . import test
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -25,7 +25,11 @@ import sys, os
 import numpy
 from . import _fitsio_wrap
 import copy
+import warnings
 import pprint
+
+class FITSRuntimeWarning(RuntimeWarning):
+    pass
 
 def cfitsio_version(asfloat=False):
     """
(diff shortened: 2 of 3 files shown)


esheldon:
that is not a warning, it is informative.

esheldon:
I mean this one
```python
print('Removing existing file')
```

rainwoodman:
I understand.

but a library shall not make assumptions on how a user utilizes stdout.
warnings is a friendlier way to issue informative notes from a library. 

As an example, if fitsio io is used in a cgi-bin script. that message would be forcely written to the webpage at the point a FITS object is created.

1. If user cares about this information it can be collected via warnings.catch_warnings (see the test case)
2. If user doesn't care about this, it can be suppressed
3. If user doesn't do anything, the warning will be issued on stderr, and stdout is untainted.

Besides, it is just replacing the call to print_function to a call to warnings.warn.

esheldon:
good point about the cgi script

esheldon:
This is way too verbose though

>>> fitsio.write('test.fits',zeros(30),clobber=True)
/home/esheldon/exports/fitsio-local/lib/python2.7/site-packages/fitsio/fitslib.py:318: FITSRuntimeWarning: Removing existing file
  warnings.warn('Removing existing file', FITSRuntimeWarning)

I don't want to see that a thousand times.  Is there a way to make this less chatty?

esheldon:
there is another problem: it only gets printed once.  The user needs to see this every time a file gets clobbered.

esheldon:
How about printing to stderr?

rainwoodman:
I added two commits to print the caller:

```
python test.py
test.py:43: FITSRuntimeWarning: Removing existing file: `/scratch1/scratchdirs/yfeng1/fitsio-ImageWrite-gMBvnU.fits`
  with fitsio.FITS(fname,'rw',clobber=True) as fits:
ok
```
Because of the added filename, each violation shall be a separate line. 

I would always deal with overwriting in the caller and decide to print to stderr or stdout there. 
From the semantics of the arguments, if 'clobber==True', it is probably safe to assume the caller  already wants to overwrite the file.

PR #37, opened 2015-04-03
Scalario

rainwoodman:
Refer to the test case I added. 

Reading in SDSS DR12 spectra file can reproduce the core dump on scalar output arrays. 
The added test case does depend on urllib2, thus we need to find a way around this.


esheldon:
hmm.. this also doesn't work if you jsut read the whole table

Sub-structs are not read correctly

esheldon:
Is this patch still needed?  Also this might have predated travis, so we don't know if it passes all the tests

esheldon:
I checked and your test case works fine in master, so I think I already addressed this in a previous commit.  I'm going to close

rainwoodman:
Is there a similar test case in master?

esheldon:
Yes, I copied your test case there

PR #38, opened 2015-04-03
Extra subset names.

rainwoodman:
The scenario is when there are two names 'Z' and 'Z_PERSON', 'Z' is displayed in **repr**  of 'Z_PERSON'.

The test case could be simplified not using the SDSS DR12 spectra, I believe.


esheldon:
Thanks for all these fixes.

A lot of this is experimental and I don't understand it all yet.  Is there a way to pull your requests into an experimental branch?  Sorry I don't know the git process for that.

The reason I ask is I'm planning to do a new release this week and I don't want to merge into master.

rainwoodman:
Maybe you can start a new branch and I file a PR against that branch.

esheldon:
Is this still needed?   looks like this now conflicts

rainwoodman:
Looks like it is fixed on master. I removed the patch but left the test case in place. I hope the test case is still ran on master.. We'll see.

rainwoodman:
Travis is green but somehow it doesn't show up here automatically.

https://travis-ci.org/rainwoodman/fitsio/builds/357170839

esheldon:
I don't think we should download data in the tests.  Can you suggest a test where you create the data internally?

rainwoodman:
I think it is no longer downloading anything any more.

fitsio/test.py
```diff
@@ -976,6 +976,31 @@ def testTableWriteRead(self):
                 #pass
                 os.remove(fname)
 
+    def testTableFormatColumnSubset(self):
```
rainwoodman: See here.

esheldon:
OK, looks good

PR #43, opened 2015-05-07
fix all False boolean for some versions of numpy.

### LINKED ISSUE #39: Problem reading boolean column from table
I have a boolean column in a FITS table (written using fitsio).  They all seem to be read back as False using fitsio but look ok when read by astropy.io.fits.

rainwoodman:
astype does the wrong thing. we want to view the same memory
chunk as an int.
Seriously, we need to develop reading numpy documentations
as a bed side hobby.

This PR fixes  #39

### DIFF
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -2344,7 +2344,7 @@ class TableHDU(HDUBase):
         If input is a fits bool, convert to numpy boolean
         """
 
-        output = (array.astype(numpy.int8) == ord('T')).astype(numpy.bool)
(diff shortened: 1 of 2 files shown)


rainwoodman:
#39 was not captured by a test case.

I am stuck writing one. eg. I am not sure why in tests.py b1scalar is translated to 'b'. Looks like it should have been '?'; but that fails 9 test cases even after the patch.

rainwoodman:
I added the test case. v0.7.9 fails the updated test. this PR passes.

Issue #2, opened 2012-08-03
Compile and link args for pure 64-bit python on OS X


esheldon:
John Parejko noted that he was using a pure 64-bit python on darwin.
The last time I had a mac we needed to compile and link with

    -arch i386 -arch x86_64

But on his version he had to remove the i386 to it to work.

Need to understand how to automatically figure this out during install.

profxj:
Just installed fitsio from the latest source (__version__='0.9.7rc1')
on my Mac laptop running OS X 10.9.5

Had to do hacks to 'succeed':

1)  Remove the -arch and -i386 from the setup.py file, i.e.

if platform.system()=='Darwin':
    extra_compile_args=['-arch','x86_64']
    extra_link_args=['-arch','x86_64']
else:

and

2)  I had to comment out the fitsio_wrap import in fitslib.py as 
I was getting a 'cannot import name' error when importing fitsio, i.e.

"""
from __future__ import with_statement, print_function
import sys, os
import numpy
#from . import _fitsio_wrap
import copy
import pprint

Things seem ok so far (have tested reading and writing FITS files).  And I just
ran the test suite (27 tests, all OK).

esheldon:
I think there must be something strange going on.  You cannot comment that line out, or the code won't work!

Did you actually install the code or just run from that directory?

profxj:
I installed it.

Here is me running the test suite from the Desktop (which knows
nothing of where I have fitstio; color me confused but a bit too novice
at Python to be of much help):

wolverine.local> cd Desktop/
/Users/xavier/Desktop
wolverine.local> ipython
Python 2.7.8 |Anaconda 2.1.0 (x86_64)| (default, Aug 21 2014, 15:21:46)
Type "copyright", "credits" or "license" for more information.

IPython 2.2.0 -- An enhanced Interactive Python.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://binstar.org
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import fitsio

In [2]: fitsio.__version__
Out[2]: ‘0.9.7rc1’

In [5]: fitsio.test.test()
testAsciiTableWriteRead (fitsio.test.TestReadWrite) ... ok
testChecksum (fitsio.test.TestReadWrite) ... ok
testExtVer (fitsio.test.TestReadWrite) ... ok
[...]

PR #46, opened 2015-05-20
removed -iarch in setup.py

cbonnett:
This should do the job


menanteau:
Not sure if this is necessary.  For DESDM we have compiled and installed without problem for 10.7-10.9 for a while now, using clang -- not CC.

esheldon:
@menanteau see issue #2   This seems to help some users.  Do you see a problem?

esheldon:
in other words, would it break the clang build?

menanteau:
Likely, as clang-compiled binaries are most of the time compiled again libraries that where also compiled in a universal way

esheldon:
OK, then we need a test for this in setup.py

PR #52, opened 2015-07-03
modified setup.py to pass numpy include dirs to _fitsio_wrap

moustakas:
With this change fitsio successfully compiles on Mac OS X 10.10.3 with (homebrew-installed) python 2.7.10.


esheldon:
We want to remove it from setup() too right?

moustakas:
include_dirs also removed from setup()

esheldon:
I guess since matt put this in his pull request #53 also, I'll just pull his

PR #54, opened 2015-07-27
Make fitsio not build the packaged cfitsio when you just want to run "clean"

joezuntz:
This change hooks the cfitsio building code in the setup.py into the standard distutils build system by sub-classing the distutils commands build and clean commands.  It makes building cfitsio part of building the extension itself.

This means that the clean command now works properly.

I haven't tried this with python 3.

Cheers!
Joe

PR #63, opened 2015-10-22
Use a subclass of build_ext to build cfitsio.

rainwoodman:
- The scripts in setup.py are refactored into a subclass of build_ext.

- CC CFLAGS ARCHIVE and RANLIB are set to the values used by python.

The second change should improve interpolibility on strange system configurations; e.g. when the system default compiler is incompatible with the one that was used to build python.

I removed the special treatment for Macs, in the hope that the python compiler should already know these command-line arguments. Could someone test if this is indeed the case? This is a WIP.


rainwoodman:
I tested this PR on a Fedora, with / without --use-system-fitsio ; both worked. (need a CFLAGS tweak --use-system-fitsio to find headers in /usr/lib64/cfitsio/)

esheldon:
This will require a lot of testing, but we should work in a branch.  Can you please submit the PR against the newly created "cfitsio_build_ext" branch?

rainwoodman:
- I am not sure if I should have filed another PR #66 
- A new subcommand for configuring and building a autotools library into the proper temp.xxxxx directory could potentially be useful in other places as well. I do not find one such thing via google yet.
- It would nice to integrate the 'clean' command in #54 .

esheldon:
wait, I already pulled 66...?

rainwoodman:
sorry the head was a bit messy as I realized there is a '-f' flag in build_ext -- and we were just trying to minic its behavior with the rm/unlink trick. 9ce6db2 enables '-f' by default.

esheldon:
Sorry, I'm confused.  You *do* want me to pull this?

rainwoodman:
I *do* want you to pull 9ce6db2  -- but I am reserved about a better solution (not setting '-f' by default) that breaks compatibility.

And I do have question about the workflow. Shall I file a new PR every time I push?

esheldon:
OK, I see now what this does.

I don't like this, because then it is not possible to see if the code was already compiled by rerunning "python setup.py build" or whatever

rainwoodman:
A follow up on subcommand: the compiler logic in distutils.commands.build_ext is sufficiently complicated; a new subcommand duplicating these logic may not be a good idea.

rainwoodman:
On https://github.com/esheldon/fitsio/pull/63#issuecomment-150667791

I thought the previous behavior there also wasn't a way to tell if the code was already compiled, since build/lib.* were purged.

rainwoodman:
5c4dfa8 should address this issue; the .a files are appended to the dependency list.

esheldon:
OK, I'll give it a try

esheldon:
OK, this was against master instead of the branch

esheldon:
This should have been against the cfitsio_build_ext branch, because none of it is tested well enough yet

rainwoodman:
Revert these changes? and move to PR #67 ?

esheldon:
Yes, I did a revert, hope it all works out

PR #64, opened 2015-10-23
Use a subclass of build_ext to build cfitsio.

rainwoodman:
Refile issue #63 for further development.


rainwoodman:
Talking about testing, we may want to throw in a travis+appveyor CI to assert a successful build on common machine types. We can probably base this off on astropy's configuration files.

esheldon:
I agree we should get travis going

PR #67, opened 2015-10-23
WIP: Cfitsio build ext

rainwoodman:
against cfitsio_build_ext.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ class build_ext_subclass(build_ext):
         self.cfitsio_dir = 'cfitsio%s' % self.cfitsio_version
         self.cfitsio_build_dir = os.path.join('build', self.cfitsio_dir)
         self.cfitsio_zlib_dir = os.path.join(self.cfitsio_build_dir,'zlib')
+
         build_ext.initialize_options(self)    
         self.link_objects = []
         self.extra_link_args = []
@@ -33,8 +34,6 @@ class build_ext_subclass(build_ext):
     def finalize_options(self):
 
         build_ext.finalize_options(self)    
-
-        self.force = True
 
         if self.use_system_fitsio:
             # Include bz2 by default?  Depends on how system cfitsio was built.
@@ -70,6 +69,11 @@ class build_ext_subclass(build_ext):
             link_objects = glob.glob(os.path.join(self.cfitsio_build_dir,'*.a'))
 
             self.compiler.set_link_objects(link_objects)
+
+            # Ultimate hack: append the .a files to the dependency list
+            # so they will be properly rebuild if cfitsio source is updated.
+            for ext in self.extensions:
+                ext.depends += link_objects
 
         # call the original build_extensions
 
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,8 @@ class build_ext_subclass(build_ext):
 
         build_ext.finalize_options(self)    
 
+        self.force = True
+
         if self.use_system_fitsio:
             # Include bz2 by default?  Depends on how system cfitsio was built.
             # FIXME: use pkg-config to tell if bz2 shall be included ?
@@ -60,12 +62,6 @@ class build_ext_subclass(build_ext):
                 self.compiler.add_library('bz2')
 
             self.compile_cfitsio()
-
-            # when using "extra_objects" in Extension, changes in the objects do *not*
-            # cause a re-link!  The only way I know is to force a recompile by removing the
-            # directory
-            for sofile in glob.glob(os.path.join('build','lib*/fitsio/*_fitsio_wrap*.so*')):
-                os.unlink(sofile)
 
             # link against the .a library in cfitsio; 
             # It should have been a 'static' library of relocatable objects (-fPIC), 
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@ from distutils.core import setup, Extension, Command
 from distutils.command.build_ext import build_ext
 
 import os
+from subprocess import Popen
 import sys
 import numpy
 import glob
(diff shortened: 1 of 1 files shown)


esheldon:
something went wrong, this is compiling the .so every time

rainwoodman:
Not here. could you print link_objects? I got
```
['build/temp.linux-x86_64-2.7/cfitsio3370patch/libcfitsio.a']
```

rainwoodman:
Does it compile libcfitsio.a every time as well?

rainwoodman:
I added travis-ci; result can be seen at

https://travis-ci.org/rainwoodman/fitsio/jobs/87121393

The current script contains calls to build and build_ext; I could not reproduce the 'always rebuild' on travis neither.

rainwoodman:
Hmm. 

'always rebuild' happens on master, as seen here:

https://travis-ci.org/rainwoodman/fitsio/jobs/87122333

I suspect you were testing on master not on cfitsio_build_ext.

esheldon:
you might be right

esheldon:
@beckermr can I enlist you to test the cfitsio_build_ext branch on your mac?

beckermr:
I cannot build with my own fitsio

```
/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'install_requires'
  warnings.warn(msg)
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: option --use-system-fitsio not recognized
```

esheldon:
looks like that option got lost

beckermr:
Not that this is a great idea any more due to various patches. However I still vote to keep it.

beckermr:
It is in the script, but for whatever reason my distutils complains

rainwoodman:
Yes option is in the script. Could be related to python version. I will take a look. 

Does it work if you run build_ext?

beckermr:
It compiles this way yes, but I cannot get it to install without compiling the packaged version.

beckermr:
This warning seems pertinent.

```/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'install_requires'
  warnings.warn(msg)```

**Comment by rainwoodman** [comment]
Thanks for the information. That warning is pertinent but not related to this PR. I figured why install fails but build_ext successes with --use-system-fitsio. It is because the proper way of supplying options to a sub-command is
```
python setup.py install build_ext --use-system-fitsio
```

I will included these to the travis CI tests. I just added two new options --system-fitsio-libdir and --system-fitsio-includedir.

beckermr:
Well that is massively confusing! I preferred the old script which would just take commands like you expect them to appear, but this is fine. This needs to be documented in the README install instructions.

rainwoodman:
I agree it was massively confusing at first -- but on second thought it makes sense since 'install' and 'build_ext' are two separate targets.

I suspect there is a clean way to propagate options within the distutils; but I will take a look. The current README doesn't seem to have instructions installing with system library.

beckermr:
Cleaning the build does not work. I ran `python setup.py clean --all` and it left the `build/cfitsio3370patch` directory.

rainwoodman:
Thanks for testing! 

This should have been fixed in the latest head of this branch. (make sure you start from an empty build directory) 

https://github.com/esheldon/fitsio/commit/a8f81c97808a9f4f025afe1d1619db073ff728a2

beckermr:
Ahh yes. Thanks! Works great.

beckermr:
the cfitsio_build_ext branch needs to be remerged into master to get this last commit (or just that last commit).

esheldon:
On 10/26/15, Matthew R. Becker <notifications@github.com> wrote:
> the cfitsio_build_ext branch needs to be remerged into master to get this
> last commit (or just that last commit).

I won't merge this until it is heavily tested.  I need more than three testers
for such a big change.  I expect in particular that messing with the build
system will break OS X where every person has a different environment.


-- 
Erin Scott Sheldon
Brookhaven National Laboratory erin dot sheldon at gmail dot com

beckermr:
Right. Not sure where we plan to get the testers from...

rainwoodman:
- We may try to add osx tests to travis-ci.

- I can talk to people with macs and ask them to test (most astronomers have a mac). The problem is what are the platforms that we are to support? How many flavors of python + c compiler are there on Mac?

- For example, I just asked @ameisner to test this on his old mac. It complained about 'testing if gcc works.... no'. It eventually boils down to gcc can't find a library named like crt10.x.x . But his compiling environment was set up 5 years ago, and he is using python from anaconda but not gcc.

esheldon:
I'll ask some previous contributors

@cbonnett @erykoff @dstndstn Any way you could test this branch?  I'm especially worried it may need adjustmentto work on all mac setups

dstndstn:
esheldon:cfitsio_build_ext builds for me, and unit tests pass.

My mac is OSX 10.8.x, 

> gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 5.1 (clang-503.0.38) (based on LLVM 3.4svn)
Target: x86_64-apple-darwin12.6.0
Thread model: posix

> python -V
Python 2.7.9

built by homebrew.

PR #68, opened 2015-10-23
Add travis-ci.

rainwoodman:
This travis-ci only tests the build and install of the package.

Enable this by flipping the switch on esheldon/fitsio after logging in at travis-ci.org. 

https://api.travis-ci.org/esheldon/fistio.svg shall contain the build status badge.

Note that the test cases in test.py are not enabled in travis. 
As there are currently 1 failure with 2 errors.


esheldon:
tests might fail if bzip2 is not present

PR #69, opened 2015-10-26
customize includedir and libdir of system cfitsio.

rainwoodman:
Note that the proper way of setting up these 'build_ext' options is via explicitly adding build_ext subcommand on the commandline, e.g.

```
python setup.py install build_ext --use-system-fitsio
``


esheldon:
I think some of these commits I already accepted in a previous PR

PR #79, opened 2016-01-29
Add ability to shrink tables

mindriot101:
This change allows the ability to shrink tables, either at the users request, or automatically when data is being overwritten.

Currently, as per cfitsio, when new rows are written to an existing table and the number of rows for the new data is less than the number of rows for the existing data, the new rows get overwritten but the old data still remains. This is quite dangerous if the user assumes the new data completely replaces the old data. This change adds functionality in three locations:
- At the C level, with `delete_rows` which deletes the rows between `start` and `end`. If `end` is not given, then delete all of the remaining rows,
- A new method on the `TableHDU` object, `_shrink_to` which deletes the last n rows leaving `target_nrows` remaining,
- A new keyword argument to `TableHDU.write`: `resize` which will automatically delete rows from the end of the table if the new data has fewer rows.

There are basic unit tests for each level, but they can certainly be tested and I haven't really thought of edge cases yet.

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -2210,6 +2210,41 @@ PyFITSObject_write_var_column(struct PyFITSObject* self, PyObject* args, PyObjec
     Py_RETURN_NONE;
 }
 
+// function to delete rows from a table
+static PyObject *
+PyFITSObject_delete_rows(struct PyFITSObject* self, PyObject* args, PyObject* kwds) {
+    int status = 0;
+    int hdunum = 0;
+    int hdutype = 0;
+    Py_ssize_t start = 0, number = 0, end = 0;
+
+    static char *kwlist[] = {
+        "hdunum", "start", "end", NULL,
+    };
+
+    if (self->fits == NULL) {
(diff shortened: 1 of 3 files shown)


esheldon:
Should we expose table resizing directly to the user?  e.g.  should we expose a `shrink` method?

We could also, in principle, have a more general `resize` method.

mindriot101:
Sorry for the delay in replying. I've updated the PR by rebasing onto master and made a couple of minor edits.

> Should we expose table resizing directly to the user? e.g. should we expose a shrink method?
> We could also, in principle, have a more general resize method.


This could be a good idea, though I can't currently think of a use case. If the user wants to remove the last few rows, it's likely they want to filter the data in a more complicated way, in which case they can read the data, perform the filtering and overwrite the data. Of course this PR makes this process more predictable.

I'm not sure it needs to be user facing. Of course the added methods here can be used if need be.

esheldon:
I decided to make a more general version of this.  I added a resize() method (as well as delete_rows method)

This is currently in the branch feature-row-modifications

please check it out

Issue #80, opened 2016-02-08
Unable to read binary table bit X type


taldcroft:
I have a binary table (from Chandra standard data products) that has a field with `TFORM2 = '263X'`.  When I try to read this I get an exception (shown below).  The file reads with `astropy.io.fits` just fine, but the table has around 900 columns and I see about 10-20 times slowdown relative to `fitsio.read()`.  For my application this is a problem.

Am I doing anything wrong, or is there just no support for the `X` type in `fitsio`?  I can provide a sample file on request.

```
In [2]: import fitsio

In [3]: t = fitsio.read('test.fits')
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-3-0557584cf732> in <module>()
----> 1 t = fitsio.read('test.fits')

/home/aldcroft/.local/lib/python2.7/site-packages/fitsio/fitslib.pyc in read(filename, ext, extver, **keys)
     92         item=_make_item(ext, extver=extver)
     93 
---> 94         data = fits[item].read(**keys)
     95         if header:
     96             h = fits[item].read_header()

/home/aldcroft/.local/lib/python2.7/site-packages/fitsio/fitslib.pyc in read(self, **keys)
   1704             data = self.read_rows(rows, **keys)
   1705         else:
-> 1706             data = self._read_all(**keys)
   1707 
   1708         return data

/home/aldcroft/.local/lib/python2.7/site-packages/fitsio/fitslib.pyc in _read_all(self, **keys)
   1725         """
   1726 
-> 1727         dtype, offsets, isvar = self.get_rec_dtype(**keys)
   1728 
   1729         w,=numpy.where(isvar == True)

/home/aldcroft/.local/lib/python2.7/site-packages/fitsio/fitslib.pyc in get_rec_dtype(self, **keys)
   2036         isvararray = numpy.zeros(len(colnums),dtype=numpy.bool)
   2037         for i,colnum in enumerate(colnums):
-> 2038             dt,isvar = self.get_rec_column_descr(colnum, vstorage)
   2039             descr.append(dt)
   2040             isvararray[i] = isvar

/home/aldcroft/.local/lib/python2.7/site-packages/fitsio/fitslib.pyc in get_rec_column_descr(self, colnum, vstorage)
   2098             See docs in read_columns
   2099         """
-> 2100         npy_type,isvar = self._get_tbl_numpy_dtype(colnum)
   2101         name = self._info['colinfo'][colnum]['name']
   2102 

[...]

PR #81, opened 2016-02-09
Add support for X (bit) binary table column type

taldcroft:
This addresses #80.

This passes the initial requirement of reading the file mentioned in #80.  

I got stuck on how to write a test for this within the existing test framework.  From what I can see that needs to write a FITS table based on an existing table.  Then I couldn't immediately figure out what to change in order to write an `X` type column.  `astropy.io.fits` punts on this and writes out `L`.  From an API perspective it seems like there is some ambiguity about how to write out a numpy `b1` dtype object (either `L` or `X`).


esheldon:
this worked, without modifying other code?

taldcroft:
It apparently works (which also surprised me).
```
In [1]: import fitsio

In [2]: t = fitsio.read('test.fits')

In [3]: t[0]
Out[3]: (567995626.477488, [False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False])

In [4]: t.dtype
Out[4]: dtype([('TIME', '>f8'), ('QUALITY', '?', (263,))])
```

taldcroft:
@esheldon - apologies, I wasn't expecting you to merge this prior to having formal tests that pass.  I was working on that this afternoon when I discovered a problem.  It turns out that while the first row has the correct values, the subsequent rows of a different column are corrupted.
```
In [1]: import fitsio
In [2]: t = fitsio.read('test_x.fits')
In [4]: from astropy.io import fits
In [5]: hdus = fits.open('test_x.fits')

In [8]: t['TIME']
Out[8]: array([  5.67995626e+008,  -1.98263180e+219,   0.00000000e+000])

In [9]: hdus[1].data['TIME']
Out[9]: array([  5.67995626e+08,   5.67995629e+08,   5.67995631e+08])
```
You can reproduce this with the `test_x.fits` file that is now in this branch.  So I confess to being a bit stuck now in terms of where to go from here.  I know only enough about the Python C API to recognize the basic structures, but I have very little practical experience debugging it, and I haven't directly used CFITSIO in many years.

esheldon:
OK, I just temporarily commented it out.  We can explore further.  Please stick to the branch.

Issue #86, opened 2016-04-23
Fitsio works (only) with unaligned data structures


olebole:
On systems like Sparc and (partly) ARM it is required that data pointers are aligned, and also on other architectures unaligned data may lead to a lower performance.
Numpy however seems to create non-aligned data structures which cause the unit tests to fail (segfault) under [Sparc64](https://buildd.debian.org/status/logs.php?pkg=python-fitsio&arch=sparc64) and [armhf](https://buildd.debian.org/status/fetch.php?pkg=python-fitsio&arch=armhf&ver=0.9.8%2Bdfsg-1&stamp=1460817790) platforms.
Aligning the data in the tests does not work yet, since this creates an unknown column type "V" for table data.

esheldon:
I think this might be straightforward, but tedious to implement.  It basically comes down to ignoring some of the fields ("V") and repacking the data before creating the table and writing.

esheldon:
Note *writing* aligned data already works fine.  For writing, references to the individual fields in the structured array are packed into a list, and in the C code I use the following to get a pointer to the data
```C
data = PyArray_GETPTR1(row);
```
this is fully aware of the alignment, strides, etc.

To make the reading also work I could restructure the C code to work in the same way. I *think* this would have minimal performance impact, but would be more general, and would support reading into aligned data as well.

esheldon:
Also note, that `arr.dtype.names` reports the names of interest, but the descriptor has the extra fields we would ignore

```python
>>> dtype=numpy.dtype([('ra','f8'),
                       ('index','i4'),
                       ('dec','f8'),
                       ('s','S7')],align=True)
>>> xa=numpy.zeros(3, dtype=dtype)


# note this looks different from unaligned, it shows
# the dtype as a dict
>>> xa
array([(0.0, 0, 0.0, ''), (0.0, 0, 0.0, ''), (0.0, 0, 0.0, '')], 
      dtype={'names':['ra','index','dec','s'], 'formats':['<f8','<i4','<f8','S7'], 'offsets':[0,8,16,24], 'itemsize':32, 'aligned':True})

# this is what we want
>>> xa.dtype.names
('ra', 'index', 'dec', 's')

# this contains stuff for padding, which we don't want
>>> xa.dtype.descr
[('ra', '<f8'),
 ('index', '<i4'),
 ('', '|V4'),
 ('dec', '<f8'),
 ('s', '|S7'),
 ('', '|V1')]
```

iainlane:
[...]

PR #87, opened 2016-04-23
Allow aligned data

olebole:
This pull request fixes partially #86, in the sense that one can now use tables that have a `dtype=numpy.dtype(..., align=True)` set. It also includes a fix for the ASCII table test to use aligned data. This is the only test that fails on the ARM platform (so, when applying this pull request, all tests for ARM [succeed](https://buildd.debian.org/status/fetch.php?pkg=python-fitsio&arch=armhf&ver=0.9.8%2Bdfsg-5&stamp=1461280405)); however for Sparc64 the other tests also result in a [segmentation fault] (https://buildd.debian.org/status/fetch.php?pkg=python-fitsio&arch=sparc64&ver=0.9.8%2Bdfsg-5&stamp=1461359715). I am not shure whether it is generally better to

- leave the decision to the user to use the `align` argument of `numpy.dtype` if he wants to use his software those platforms (which would make the software potentially unportable, with hard to find bugs later), or
- re-align the data in fitsio (this package), or
- handle this case in libcfitsio.
Any ideas?

### DIFF
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -3426,6 +3426,8 @@ def array2tabledef(data, table_type='binary'):
             # the same type as the first
             name=d[0]
             form, dim = npy_obj2fits(data,name)
+        elif npy_dtype[0] == "V":
+            continue
         else:
             name, form, dim = npy2fits(d,table_type=table_type)
 
--- a/fitsio/test.py
+++ b/fitsio/test.py
@@ -182,7 +182,8 @@ class TestReadWrite(unittest.TestCase):
                 ('f8scalar','f8'),
                 ('Sscalar',Sdtype)]
         nrows=4
-        adata=numpy.zeros(nrows, dtype=adtype)
+        # For armhf and sparc64, we need alignment here:
+        adata=numpy.zeros(nrows, dtype=numpy.dtype(adtype, align=True))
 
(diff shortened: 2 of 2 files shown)


esheldon:
I think the align=True keyword is very recent in numpy.

Should probably wrap that in a try: except: to make sure it will work on older numpy

```python
try:
    tdt=numpy.dtype(adtype, align=True)
except:
    tdt=numpy.dtype(adtype)
adata=numpy.zeros(nrows, dtype=tdt)
```

olebole:
The `align` parameter is in `numpy.dtype` at least since [version 1.3](http://docs.scipy.org/doc/numpy-1.3.x/reference/generated/numpy.dtype.html?highlight=align), which was released in 2009. I guess that other things break as well if you use such an old version.

esheldon:
On Apr 25, 2016 3:02 AM, "Ole Streicher" <notifications@github.com> wrote:
>
> The align parameter is in numpy.dtype at least since version 1.3, which
was released in 2009. I guess that other things break as well if you use
such an old version.

Maybe, but not that i'm aware.  The code base was originally developed on
pre 1.0 numpy

olebole:
If you need compatibility with pre-2009 numpy, I can still add your patch. Shall I?

esheldon:
yes, please

PR #89, opened 2016-05-20
Adding "byte_offsets" property to HDU object.

demitri:
Adds `byte_offsets` property to HDU object, implementing the cfitsio routine [fits_get_hduaddrll](https://heasarc.gsfc.nasa.gov/docs/software/fitsio/c/c_user/node65.html). This returns the byte offset for the header start, data start, and data end of the HDU.


esheldon:
sorry I didn't get to this yet.

I have a slightly different implementation in mind. This is a property of an HDU, so I think the right interface would be

```python
with fitsio.FITS(filename) as fits:
    offset=fits[ext].get_byte_offset()
    # or maybe
    offset=fits[ext].get_hduaddr()
```

in any case it is straightforward to add this property to the objects.

esheldon:
I added this to the information dictionary, which you can get with the get_info() method.  This is in master.

```
info=fits[ext].get_info()
info['header_start']
info['data_start']
info['data_end']

# You can also get this info in a small dictionary with just those specific entries
offsets=fits[ext].get_offsets()

```

demitri:
Thanks for pulling that in. I later saw the HDU object and moved the property there in my implementation. I saw the info dictionary, but didn't want to calculate the values until the user actually needed them. I have millions of FITS file to process - I prefer lazy loading to avoid any calculations I don't need. Anyway, thanks!

esheldon:
can we close this request?

demitri:
Yes, I'm completely happy with it, thanks! I might request a new release when you have a chance; it's easier to ask others to update to the "current version" than download and install from trunk here.

esheldon:
just tagged v0.9.9

I'll upload to pypi today

demitri:
Thanks!

esheldon:
released on pypi

Issue #90, opened 2016-07-02
write_keys fails to write comments


kadrlica:
With fitsio 0.9.7 I've been trying to copy a header from one hdu to another and I've noticed that the comments are not preserved. Looking [here](https://github.com/esheldon/fitsio/blob/2b1e8fafd4dfaca4e7f6741cf6c83d89658f8da0/fitsio/fitslib.py#L1361-L1367), I think the problem is that the `value` of a `COMMENT` record is blank. For example:

```
{'card': 'COMMENT 160627_28/iff1025c8.fits ut160627_28/iff1026c8.fits',
 'comment': '160627_28/iff1025c8.fits ut160627_28/iff1026c8.fits',
 'name': 'COMMENT',
 'value': '',
 'value_orig': ''}
```

esheldon:
did you also try git master?



On 7/2/16, Alex Drlica-Wagner <notifications@github.com> wrote:
> With fitsio 0.9.7 I've been trying to copy a header from one hdu to another
> and I've noticed that the comments are not preserved. Looking
> [here](https://github.com/esheldon/fitsio/blob/2b1e8fafd4dfaca4e7f6741cf6c83d89658f8da0/fitsio/fitslib.py#L1361-L1367),
> I think the problem is that the `value` of a `COMMENT` record is blank. For
> example:
>
> ```
> {'card': 'COMMENT 160627_28/iff1025c8.fits ut160627_28/iff1026c8.fits',
>  'comment': '160627_28/iff1025c8.fits ut160627_28/iff1026c8.fits',
>  'name': 'COMMENT',
>  'value': '',
>  'value_orig': ''}
> ```
>
> ---
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/esheldon/fitsio/issues/90


-- 
Erin Scott Sheldon
Brookhaven National Laboratory erin dot sheldon at gmail dot com

kadrlica:
I just tried with the current master and I have the same issue:
```
> git describe
v0.9.8rc2-19-g2b1e8fa

> python
>>> import fitsio
>>> fitsio.__version__
'0.9.8'
>>> fits = fitsio.FITS('bpm/iffbpmc1.fits')
>>> fits[0].read_header()

SIMPLE  =                    T / file does conform to FITS standard
BITPIX  =                   16 / number of bits per data pixel
NAXIS   =                    2 / number of data axes
[...]

PR #91, opened 2016-07-02
Copy COMMENT and HEADER fields with write_keys

kadrlica:
This might be a fix for issue #90, but I haven't tested it very throughly. It appears to work for the specific case that was bothering me (i.e., copying COMMENT strings), but I'm not sure about HISTORY.


esheldon:
this is superseded by the fix I just pushed to master

Issue #100, opened 2016-09-13
Writing header keys in a particular order


rainwoodman:
Hi,

It was mentioned at https://github.com/legacysurvey/legacypipe/issues/108#issuecomment-246568380
that the ordering of keys in a file directly written with write_header is not sorted. Is there a way to specify the ordering of keys in a FITSHDR? The documentation seems to be unclear about this and I could not find any portion of the API suggesting an easy way of doing so.

Thanks,

rainwoodman:
I figured it is probably done by passing in a list of dict. If this is true it may be worthy to explictly mention this in the docstring.

esheldon:
I posted a comment there, repeating below.  I agree this should be more clearly documented.

If you create the header as a plain dict, the keys will not remember any ordering.

The default way to get ordering is to create a FITSHDR object directly:
```python
h=fitsio.FITSHDR()
h.add_record({'name':'stuff', 'value':3.0,'comment':'this is a comment'})
h.add_record({'name':'greet', 'value':'hello'})
h.add_record({'name':'RA', 'value':'215.0'})
```

an OrderedDict might also work, but I haven't tried it.

rainwoodman:
Closed via #101

PR #101, opened 2016-09-13
Mention orderness in docstrings related to header.

rainwoodman:


rainwoodman:
This will sort of close #100. Unless we want to mention OrderedDict.

PR #105, opened 2016-11-21
read_raw

dstndstn:
This PR adds a read_raw() method to the fitsio_pywrap.c FITS class.  It flushes data, copies the full file contents, and returns it as a Python string.

This should be particularly useful for "mem://" files.

This is motivated by a kind of weird use case where we want to catch filesystem corruption, so we do:

```
F = fitsio.FITS('mem://', 'rw')
F.write( ... )
rawdata = F.read_raw()
sha = hashlib.sha1()
sha.update(rawdata)
# remember sha.hexdigest()
f=open('F.fits', 'w')
f.write(rawdata)
f.close()
# Run sha1sum F.fits, compare against sha.hexdigest().
```

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -3525,7 +3525,7 @@ PyFITSObject_read_raw(struct PyFITSObject* self, PyObject* args) {
     // Allocate buffer for string
     sz = FITS->filesize;
     // Create python string object of requested size, unitialized
-    stringobj = PyString_FromStringAndSize(NULL, sz);
+    stringobj = PyBytes_FromStringAndSize(NULL, sz);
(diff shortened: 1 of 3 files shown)


esheldon:
Thanks Dustin.  Let me know when you think this is ready for review

dstndstn:
I *thought* it was ready when I submitted the PR, but then Travis got mad at me (correctly) for Py3 problems.  That's fixed, so I think it's good to go now.  I also checked that I got the memory management right.  Thanks!

dstndstn:
Ready for review :)

esheldon:
this looks good to me, no need for a full review

dstndstn:
Thank you!

PR #115, opened 2017-03-31
Support fits_open_memfile/fits_create_memfile

irasnyd:
It is occasionally useful to be able to use fitsio on files which you
have already read into memory through other means. The cfitsio library
supports this with fits_open_memfile/fits_create_memfile. Support is now
present for this method of operation.

This feature was developed with the use case of a thumbnail server in
mind. In our implementation, FITS files are received over HTTPS from the
network, immediately converted into a JPEG, and then sent out over HTTPS
to the client. By keeping everything in memory, the requests are
processed a little bit faster, and we don't need to worry about filling
up our disks with temporary files over time.

The implementation is somewhat suboptimal, because I could not find a
way in the Python interface that lets me use realloc() on a bytearray
(or similar Python object). Therefore, I had to create a copy of the
bytearray and use it instead, only copying back into the bytearray when
calling the close() method.

If we change this feature so that the only supported mode is read-only,
then this suboptimal behavior can go away (it becomes zero copy). **Zero
copy mode breaks the read_raw() accessor (SEGV in memcpy()). If I remove
zero copy mode and always copy, then everything I've tested works fine.**

**However, as an initial draft, I thought it would be beneficial to show
that we can do both read-only and read-write modes.**

Signed-off-by: Ira W. Snyder <isnyder@lco.global>

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,37 @@ f[1].get_vstorage()         # for tables, storage mechanism for variable
 f[1].lower           # If True, lower case colnames on output
 f[1].upper           # If True, upper case colnames on output
 f[1].case_sensitive  # if True, names are matched case sensitive
+
+# Open a file in memory, using data you have already read in
+# through other alternative means (from the network, etc.).
+# The filename is ignored!
+bytes = bytearray(data_in_memory)
+f = FITS('inmemory.fits', 'r', bytes=bytes)
+f[0].read_header()
+
+# You can also create files directly in memory without writing them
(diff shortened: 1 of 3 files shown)


esheldon:
thanks, this looks interesting.

I'm going on vacation and probably won't be able to look in detail for a week or more.

One thing I wonder is if it would be better to have a new class `FITSBytes` rather than a new keyword. Not sure.

irasnyd:
Superceded by PR #122.

PR #116, opened 2017-04-03
Support fits_open_memfile/fits_create_memfile

irasnyd:
This is an alternate version of my previous pull request.

Rather than adding a parameter to the FITS class, we instead
create a FITSMemFile class which inherits almost everything from
the FITS class. This makes both implementations share the vast
majority of code, but keeps the differences between each implementation
separate and makes them easier to understand.

I also added a method to the test suite to make sure that the FITSMemFile
class is working as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,18 @@ f[1].get_vstorage()         # for tables, storage mechanism for variable
 f[1].lower           # If True, lower case colnames on output
 f[1].upper           # If True, upper case colnames on output
 f[1].case_sensitive  # if True, names are matched case sensitive
+
+# Open a file in memory, using data you have already read in
+# through other alternative means (from the network, etc.).
+with FITSMemFile(bytearray(data_in_memory), 'r') as f:
+    f[0].read_header()
+
+# You can also create files directly in memory without writing them
+# to disk. This is useful if you are going to immediately send them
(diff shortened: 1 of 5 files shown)


irasnyd:
I would be perfectly happy to remove all of the code for read-write mode and make this class only
supported for read-only operation. It would significantly reduce the complexity of the implementation.

The read-only mode is sufficient for my intended use case.

The FITS class already supports creating a read-write file in memory using the special filename "mem://".
This method only supports creating a new file (rather than opening an existing file which is already in memory and changing the data within). You can see an existing usage in the test suite itself:

* https://github.com/esheldon/fitsio/blob/master/fitsio/test.py#L1628

Let me know if you think support read-only mode is a better way forwards.

Thanks!

esheldon:
I'm sorry I let this slip.  I'll try to have a look soon.

esheldon:
If you are happy with read only, and the code is much simpler, we should do that

esheldon:
also do you mind sending the PR against a branch?  Ive created the branch memfile

irasnyd:
This PR has been superceded by PR #122. It is the same changes, except that the read-write portion is stripped out, leaving only the read-only portion behind. This makes the changes to the C code very minimal.

Issue #118, opened 2017-04-27
Implement `CONTINUE` convention


kadrlica:
This follows from the discussion in #117. The `cfitsio` library has functions implemented for reading, writing and manipulating long strings following the `CONTINUE` convention described here:

https://fits.gsfc.nasa.gov/registry/continue_keyword.html

Specifically, we would want to interface with the `cfitsio` functions like 
[`fits_read_key_longstr`](https://github.com/kadrlica/fitsio/blob/master/cfitsio3370patch/longnam.h#L138)
[`fits_write_key_longstr`](https://github.com/kadrlica/fitsio/blob/master/cfitsio3370patch/longnam.h#L82)
to manipulate long sting headers.

Since this could be a significant change to how header cards are parsed (and a non-zero investment of time), I have created a new issue to track this development.

kadrlica:
It appears that LSST is also using the `CONTINUE` convention. This probably motivates us interested parties (@kadrlica @erykoff @rmjarvis) to do a proper implementation sometime soon.

buchbend:
For our in-house created fits files for the upGREAT instrument on SOFIA we are also using the CONTINUE convention and would be very interested in an implementation in ´fitsio´. We would greatly benefit from the increased performance over ´astropy.io.fits´, which we are currently using to parse the fits-headers.  Unfortunately, I did not have a great idea either on how to implement this myself easily, when checking the source code of fitsio.

rmjarvis:
A start would be to submit a test that currently fails.  It's not clear to me what kind of `CONTINUE` usage is currently causing anyone problems.  I haven't run into any trouble with LSST images yet.  Without a clear description of what isn't working, it's hard to know what kind of change is required.

buchbend:
Sorry for the vague information from my side. I am using Ubuntu 18.04 with python 3.6.5. I have two issues with the current state:

### Keywords with CONTINUE are not read as long strings

I looked at the fitsio test suite and found an existing test related to CONTINUE. So far this test only checks that the CONTINUE keys are seen. A test for an implementation of the CONTINUE convention would be to check that the value of the header keyword followed by CONTINUEs is actually the long string that is encoded in the subsequent entries. For the existing Test ` testHeaderContinue` (current master branch `38df6f0`) the following snippet added in line `371` would test a sucessfull read of the CONTINUE convention:

~~~ .py
with fitsio.FITS(fname) as fits:
    rh = fits[0].read_header()
[...]

PR #119, opened 2017-04-28
Treat CONTINUE card like COMMENT

### LINKED ISSUE #117: Change to CONTINUE behavior?
I recently updated from `v0.9.8rc2` to `v0.9.11` and I've noticed a difference in how `CONTINUE` cards are being dealt with in headers.

I have a header that looks something like this:
```
OBSTYPE = 'object  '           / Observation type                               
CAMSHUT = 'Open    '           / Camera shutter at exposure start              
PROGRAM = 'Precision Astrometry with DECam: A dynamical study of nearby &'      
CONTINUE  'clusters&'                                                          
CONTINUE  '&' / Current observing orogram                                      
OBSERVER= 'DJN, DJJ'           / Observer name(s)                              
PROPOSER= 'BertinE '           / Proposal Principle Investigator               
PROPID  = '2012B-3001'         / Proposal ID                                    
```

In `v0.9.8rc2` this header read without complaint, though it didn't fill the value of the `PROGRAM` card fully:
```
>>> hdr['PROGRAM'] 
'Precision Astrometry with DECam: A dynamical study of nearby &'
>>> hdr['CONTINUE']
''
```

When I try to read this header with `v0.9.11` I get the following exception
```
/cvmfs/des.opensciencegrid.org/fnal/anaconda2/envs/des17a/lib/python2.7/site-packages/fitsio/fitslib.pyc in set_card(self, card_string)
   4300 
   4301                 if front=='CONTINU':
-> 4302                     raise ValueError("CONTINUE not supported")
   4303 
   4304                 if front=='HISTORY':

ValueError: CONTINUE not supported
```

In contrast, `pyfits` (through astropy) seems to be ok with this header and gives:
```
>>> hdr['PROGRAM']
'Precision Astrometry with DECam: A dynamical study of nearby clusters'
```

I can understand if `fitsio` doesn't deal with `CONTINUE` cards, but is there a way for me to still read this header? My specific suggestion would to print a warning rather than raising an exception.

kadrlica:
This PR closes #117 by treating `CONTINUE` cards like `COMMENT` cards. This is not a full implementation of the `CONTINUE` convention documented [here](https://fits.gsfc.nasa.gov/registry/continue_keyword.html), which is left to be addressed by #118.

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -2456,6 +2456,74 @@ PyFITSObject_write_history(struct PyFITSObject* self, PyObject* args) {
 
     Py_RETURN_NONE;
 }
+
+//   ADW: Adapted from ffpcom and ffphis in putkey.c
+int fits_write_continue( fitsfile *fptr,      /* I - FITS file pointer  */
+                         const char *cont,    /* I - continue string    */
+                         int   *status)       /* IO - error status      */
+/*
+  Write 1 or more CONTINUE keywords.  If the history string is too
+  long to fit on a single keyword (72 chars) then it will automatically
+  be continued on multiple CONTINUE keywords.
+*/
+{
+    char card[FLEN_CARD];
+    int len, ii;
+
+    if (*status > 0)           /* inherit input status value if > 0 */
+        return(*status);
+
+    len = strlen(cont);
+    ii = 0;
+
+    for (; len > 0; len -= 72)
+    {
+        strcpy(card, "CONTINUE");
+        strncat(card, &cont[ii], 72);
+        ffprec(fptr, card, status);
+        ii += 72;
+    }
+
+    return(*status);
+}
+
+// let python do the conversions
+static PyObject *
+PyFITSObject_write_continue(struct PyFITSObject* self, PyObject* args) {
+    int status=0;
+    int hdunum=0;
+    int hdutype=0;
+
+    char* value=NULL;
+ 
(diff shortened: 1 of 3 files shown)


esheldon:
Alex, it looks like pull request is against master rather than continuekey

kadrlica:
Ok, I guess that's not what you wanted. I changed the base to `continuekey`.

esheldon:
I have a couple of comments.

- I think we should not expose `write_continue` directly, instead maybe have it called `_write_continue`.  I think this might go away when we learn how to do this properly, so we should hide it for now.  Don't want people to start using it or relying on it.

- is there any reason not to always use the long version for writing?

kadrlica:
I agree completely with the first point (I'll commit a change).

For the second point, I just didn't want to mess with anything that was working. There is one issue that I can see for implementing long strings. Currently we don't support reading or writing continue statements (they are treated like comment cards and are not updated if the header card value is changed). If we implemented the long string write, then we would be allowing people to write long strings, but not read them back in. This would probably lead to some confusion, and it is probably better to just raises an exception for long string values until long string support is fully implemented.

esheldon:
ok

another question: have you run this in production yet?

kadrlica:
I've only run in testing.

esheldon:
OK, I'll test this today

esheldon:
I haven't forgotten about this, I've just been busy. I'll try to get to it soon.

kadrlica:
Would it be possible to get a `post` or `rc` tag for installation versioning purposes?

esheldon:
tagged v0.9.12rc1

PR #123, opened 2017-06-19
Bitcols

erykoff:
Support for bitcol (TBIT) arrays.

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -812,12 +812,16 @@ PyFITSObject_get_hdu_info(struct PyFITSObject* self, PyObject* args) {
 
 // this is the parameter that goes in the type for fits_write_col
 static int 
-npy_to_fits_table_type(int npy_dtype) {
+npy_to_fits_table_type(int npy_dtype, int write_bitcols) {
 
     char mess[255];
     switch (npy_dtype) {
         case NPY_BOOL:
-            return TLOGICAL;
+            if (write_bitcols) {
+                return TBIT;
+            } else {
+                return TLOGICAL;
+            }
         case NPY_UINT8:
             return TBYTE;
         case NPY_INT8:
@@ -1737,6 +1741,7 @@ PyFITSObject_write_column(struct PyFITSObject* self, PyObject* args, PyObject* k
     int hdunum=0;
     int hdutype=0;
     int colnum=0;
+    int write_bitcols=0;
     PyObject* array=NULL;
 
     void* data=NULL;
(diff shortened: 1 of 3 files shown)


esheldon:
Cool, thanks for this.  Have you checked if there is any performance hit for normal reads, without bit cols?  You could run the unit tests with and without.

We need a unit test for this.  That would require adding write support as well.  We could put that off if you don't have an immediate idea how to implement that.

erykoff:
runtime on the test looks identical within noise with a few tries on my machine.  However, if I run fitsio.test.test() two times in a row it fails with `FAIL: testNonStandardKeyValue (fitsio.test.TestWarnings)`
I actually don't think that adding write support would be too hard...let me see...

esheldon:
weird that error

erykoff:
There is a problem in writing that there's an ambiguity between an array of bools (L) and a bit array (X). Another keyword to say if you want the bit array? (This is a more compact storage on disk but same in memory in the current defaults).

-Eli.

On Jun 19, 2017, at 12:31 PM, Erin Sheldon <notifications@github.com<mailto:notifications@github.com>> wrote:


weird that error

erykoff:
Write support and testing now in.  Bitcols will only be written if you do "write_bitcols".  I just realized I haven't updated the docs... but let me know if this looks like a reasonable interface and I'll update that.  (An option would be to write bitcols instead of large boolean arrays always, which would simplify the code but could break somebody's code)

esheldon:
Yes, I'm not sure.  There is some penalty to dealing with these (e.g. reading the whole file will be slower).  So maybe that can be the deciding factor.

erykoff:
Not sure if reading the whole file will be slower.  The files will be a lot smaller...I don't know which factor wins.  Anyway, this would be the most backwards-compatible method, so maybe stick with it?

esheldon:
I think you are correct.  This is a situation where I'd like to poll  users, but that isn't really practical.

erykoff:
Okay, should be good to go (once it passes ci).  Speed is the same as before on the tests, plus the epsilon for the additional write/read bitcols tests (which pass!)

esheldon:
what about appending to a table?  Does the append use bitcols when the write used bitcols?

erykoff:
Yup.  Just added a testTableBitcolAppend test and confirmed that it uses the same format as what was written into the file at the initial write.

erykoff:
Meanwhile, I noticed that the test leaves behind TestWarning, HeaderContinue, HeaderWrite, and readraw files in /tmp.  But I don't want to fix this on this branch.

esheldon:
@taldcroft bitcols have been implemented by @erykoff on this branch.

Could you have a look and see if this solves your issue?

esheldon:
@taldcroft I just merged into branch bitcols

esheldon:
I think this is the last big TODO before going to version 1.0 (no one seems to care about HDU groups).

so before merging I think it is worthwhile to wait to hear from aldcroft, since he is the other person who was interested in this.

erykoff:
Okay.  Meanwhile, I just checked (only now...) and it seems that astropy.io.fits is able to pack the bits into an array of bytes (u1), but somehow that gets interpreted as a numpy bool array.  I assume it's more memory efficient, but to support this I think it would be more complicated ... plus I'd have to figure out how that magic happens.

taldcroft:
Sorry, been busy with astropy release and other stuff.  Here is an example file which I need to read that contains a packed bitcol (the QUANTITY column):

 http://cxc.harvard.edu/mta/ASPECT/tmp/cxc_eng_data.fits.gz

Can you see if that reads in OK?  I think most (if not all) of the values should be False.

erykoff:
@taldcroft I just looked at that file, and the QUALITY (not QUANTITY -- we're all about QUALITY over QUANTITY here...) column is all False and reads in just fine.  I checked and it's also all False with astropy.io.fits.

taldcroft:
Excellent, thanks!  I have a speed-critical application (reading thousands of FITS files every day) so when I have a chance I will bench fitsio and see if it makes sense to switch from astropy.io.fits.

Issue #121, opened 2017-05-30
travis link broken


esheldon:
Can someone who understands travis integration look into this?

esheldon:
I think this was put in place by @rainwoodman 

Do you know how to fix the link in the README?

PR #128, opened 2017-06-28
fix travis link.

rainwoodman:


rainwoodman:
This is intended to fix #121

Issue #130, opened 2017-07-24
Corrupted CONTINUE Cards


kadrlica:
I've been using the stop-gap fix in `v0.9.12rc1` to treat `CONTINUE` keys like comments (#117 and #119). So far things have been working well (over ~100k images processed). However, I just hit this error:
```
ValueError: bad card 'CONTINUE= '        '           /   '&' / Current observing orogram'.  CONTINUE not supported
```

Below is a snippet of the offending header:
```
CAMSHUT = 'Open    '           / Camera shutter at exposure start               
PROGRAM = 'Catching Supernovae in the Act with KEGS (Kepler Extra-Galactic &'   
CONTINUE= '        '           /   '&' / Current observing orogram              
OBSERVER= 'DJJ     '           / Observer name(s)                               
```

Something is amiss here. The [FITS convention](https://fits.gsfc.nasa.gov/registry/continue_keyword.html) states that keywords should "have the name  CONTINUE in bytes 1 through 8, and have space characters in bytes 9 and 10" (i.e., no "="). The header above doesn't obey this convention, and thus can be considered corrupt.

Now, we could just claim this is a corrupt header and wash our hands of the matter. However, I'm pretty sure that this header got corrupted by being processed by an older version of `fitsio` that was not treating `CONTINUE` cards correctly. The original header before processing:
```
CAMSHUT = 'Open    '           / Camera shutter at exposure start
PROGRAM = 'Catching Supernovae in the Act with KEGS (Kepler Extra-Galactic &'
CONTINUE  'Survey) and DECam&'
CONTINUE  '&' / Current observing orogram
OBSERVER= 'DJJ     '           / Observer name(s)
```

This type of corruption is consistent with how older versions of `fitsio` dealt with `CONTINUE` cards. Even if the corrupted header does not strictly conform to the FITS convention, it would be nice if we could at least read it with newer versions of `fitsio`.

The change would be relatively minor, just adding `or front=='CONTINU'` [here](https://github.com/esheldon/fitsio/blob/cf0cfec03b9971eeb733a7e85ddb16ab0a0bb015/fitsio/fitslib.py#L4385). However, this would break with the standard convention, which states that CONTINUE will not be followed by "=".

Thoughts from @esheldon?

esheldon:
Actually it should have been read as a normal key, but we still have this in place

[...]

PR #133, opened 2017-09-18
Add test of real DES header with CONTINUE= and fix to work.

rmjarvis:
I was running into some errors when reading DES Y3 fits files that had "CONTINUE=" lines (rather than CONTINUE with out the =).  I don't know if this is FITS-standard-conforming or not, but it seems like it is worth fitsio supporting this syntax.

### DIFF
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -4382,7 +4382,7 @@ class FITSCard(FITSRecord):
             self._check_len()
 
             front=card_string[0:7]
-            if (not self.has_equals() or front=='COMMENT' or front=='HISTORY'):
+            if (not self.has_equals() or front in ['COMMENT', 'HISTORY', 'CONTINU']):
 
                 #if front=='CONTINU':
                 #    raise ValueError("CONTINUE not supported")
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -4384,9 +4384,6 @@ class FITSCard(FITSRecord):
             front=card_string[0:7]
             if (not self.has_equals() or front in ['COMMENT', 'HISTORY', 'CONTINU']):
 
-                #if front=='CONTINU':
-                #    raise ValueError("CONTINUE not supported")
-
                 if front=='HISTORY':
                     self._set_as_history()
                 elif front=='CONTINU':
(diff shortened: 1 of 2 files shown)


rmjarvis:
Any chance we could get this merged into master and onto an official tag soon?  For provenance purposes, it would be nice if my runs of piff for DES Y3 used the main-line fitsio release version rather than my own custom fix.  (Also, maybe time to bump up to version 1.0?...)

esheldon:
@kadrlica how does this fit in with your work on continue?

kadrlica:
I like the test that @rmjarvis added, it should definitely be there. I think that the change to `fitslib.py` is ok, but I don't think it address the full suite of problems identified in #130.

rmjarvis:
I added the warning discussed in #130.  And a test that it was properly emitted.  I think this is ready for review.

esheldon:
I know @kadrlica is busy with job stuff, but should we go ahead and merge this?  The DES folks are asking for it so they can do some production runs

rmjarvis:
I would be happy to have this merged, if you're happy with it.  That would let me use the mainline fitsio for my Piff runs, which would be preferable.  (I'm currently using this branch.)

esheldon:
OK, I'll wait a day or so for a response from Alex; we can always back it out later if we don't like it

kadrlica:
Yep, I'm fine with merging this (I think the fact that @rmjarvis and I did made the changes in the same way independently is good enough as a review). We just shouldn't forget about the remaining issues, so I suggest that we leave #130 open.

esheldon:
pypi changed how they work and fitsio can no longer be pushed to there.  They now require a manifest, among other changes.

I could really use some help making this pypi ready.  The issue is that in order to test if it works you have to create a new version.  And if it fails, that version cannot be used again. So you end up with versions like 1.0.0.12341 after 12341 different tries to meet their criteria :)

rainwoodman:
OK. I'll file a PR to kill all warnings about packaging. BTW, isn't there a test version of pypi for this sort of things?

rmjarvis:
I've never used it.  But yes: https://packaging.python.org/guides/using-testpypi/

PR #135, opened 2017-09-21
Link against libm.

rainwoodman:
Because fitsio uses it sinh, tanh, and similar functions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -101,7 +101,10 @@ class build_ext_subclass(build_ext):
             # FIXME: use pkg-config to tell if bz2 shall be included ?
             self.compiler.add_library('cfitsio')
             pass
-
+        
+        # fitsio requires libm as well.
+        self.compiler.add_library('m')
+        
         # call the original build_extensions
 
         build_ext.build_extensions(self)


esheldon:
why is this needed now if it was not needed previously?

rainwoodman:
Here is my picture of the situation:

1. The proper way -lm shall be added to the linker. It declares a dependency on libm, which it should. In addition, only if -lm is added to the linker,  ld is able to open libm.so and find the symbols, get their versions, and create a corresponding version segments for these math symbols. It is more reliable ABI (because libm symbols change meanings from version to version, in principle.)

2. In practice on Linux-64 it usually won't break if -lm is missing. without -lm added to the linker, ld will created unversioned symbols in the table, not even knowing the so file depends on libm.so. They are still correctly resolved when the extension is loaded, only because Python has already load libm before the extension is loaded.

This issue emerged when I tried to pin down the version of function used by the extension to an earlier version of GLIBC (to allow compiling the conda package on travis-ci but running it on RocksCluster 6.x and older crays). When I did that the linker reported there is no segment for versioning the libm symbols in the fitsio extension so file. That's how I figured the extension module forgot to link to libm.

I can keep this patch in the conda recipe, but I feel it is a good idea to actually do the 'right thing' instead of just do the 'worked for me' thing in the package.

PR #139, opened 2017-11-29
Added ignore_empty option to allow the writing of empty (no data) HDUs 0

menanteau:


esheldon:
I was thinking it could be even simpler, turn off the check if empty is ok.

Would that also work?

esheldon:
Also we need a unit test for this

PR #140, opened 2017-11-29
Added ignore_empty option to allow the writing of empty (no data) HDUs 0

menanteau:
Added kwarg to allow ignoring call to _ensure_empty_image_ok() function.

### DIFF
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -420,6 +420,7 @@ class FITS(object):
               compress=None, tile_dims=None,
               header=None,
               names=None,
+              ignore_empty=False,
               table_type='binary', write_bitcols=False, **keys):
         """
         Write the data to a new HDU.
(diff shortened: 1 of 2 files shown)


esheldon:
Can we make it even simpler, just remove the `self._ensure_empty_image_ok()` altogether?

Or maybe we could make this an option on construction of the FITS object, defaulting to `allow_empty=True` ?

menanteau:
We would remove `self._ensure_empty_image_ok() altogether`, but I guess the function was there for a reason

esheldon:
It was there so that people would not write empty headers and then try to `fitsio.read(fname, ext=ext)` but maybe that isn't reallyl necessary

esheldon:
actually I see that this was part of the write function, but I think it would be better in the constructor

menanteau:
Sure, do you want to make the changes yourself?

esheldon:
I think it would be easier for you to do it, then just push changes to your branch.  It will automatically update I think

menanteau:
Just made the changes, but I'm not sure they were automatically updated.

PR #144, opened 2017-12-17
Protect numpy import in setup.py

### LINKED ISSUE #143: import numpy in setup.py
The unconditional `import numpy` in setup.py is causing my [readthedocs builds to fail](https://readthedocs.org/projects/bossdata/builds/6426290/).  I can probably work around this, but do you really want to require that numpy is installed just to run `python setup.py egg_info` on your pkg?  Perhaps you could wrap your `import numpy` in a try or a function that is only called when actually needed?

dkirkby:
This fixes #143 and allows readthedocs to cope with a fitsio dependency.

Sorry for the noisy diff (my editor automatically trims any invisible trailing whitespace).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,16 +5,15 @@ from distutils.command.build_ext import build_ext
 import os
 from subprocess import Popen
 import sys
-import numpy
 import glob
 import shutil
 import platform
 
 class build_ext_subclass(build_ext):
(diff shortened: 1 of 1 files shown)


esheldon:
Thanks for the pull request.

However, this is too noisy, it is very hard to tell what is happening in the commit.

I will just do this separately.

esheldon:
I pushed the change to master

PR #147, opened 2018-01-12
Upgrade to cfitsio 3420

rmjarvis:
I recently ran into trouble installing fitsio on my Mac when using gcc rather than clang (via conda install gcc -- guess why I was experimenting with that...).  The problem seemed to be with the `-arch i386` flag that the cfitsio configure added to the CFLAGS.  I checked and the latest cfitsio version 3420 already had the fix in.  

So I made a branch on my fork of fitsio and updated it to use cifitsio 3420, and I was going to make a pull request from that.  However, I noticed that you had a nearly identical branch on this repo, also called cfitsio3420.   The only difference is that I wasn't bold enough to remove the two old versions.

Anyway, my request is just to please merge this into master to fix the build problem for the particular combination of anaconda on Mac using the conda's gcc as the compiler.


rmjarvis:
I'm afraid I don't understand the error travis is reporting about 
```
ImportError: /home/travis/fitsio-temp-install/lib/python2.6/site-packages/fitsio/_fitsio_wrap.so: undefined symbol: curl_global_cleanup
```
So I think I need to lean on you to diagnose that error.

esheldon:
Yes, I'm working on this in a different branch.

esheldon:
I'm going to close this, eli is working on the latest cfitsio (with some patches he has applied) in #154

PR #154, opened 2018-03-20
Upgrade to latest cfitsio 3.430, with patches

erykoff:
Now with u1 and u2 support for PLIO compression, and GZIP_2 maskplane bugs fixed.


esheldon:
@erykoff looks like the travis tests are failing

erykoff:
Okay, I see that it's failing on the `--use-system-fitsio` tests, because the system fitsio doesn't support PLIO compression of unsigned ints.  What is the path forward?

esheldon:
I think the travis script needs to install the new cfitsio

erykoff:
Ah, okay, I see that now.  I'll add that.

erykoff:
...And this failed as well, because the method of checking whether we need `-lcurl` looks at the Makefile, which doesn't exist when using external cfitsio.  Can we just do `-lcurl` if curl is detected at all, and how do we do that in python?

esheldon:
I don't know the answers

erykoff:
Okay, one can look at the object table with `objdump -f` to see if `curl` or `bzip2` routines are in the system-provided `libcfitsio`.  I don't know if this is exactly kosher, but it would work.  
The next issue is that the HCOMPRESS test with unsigned u2 ints is going to fail with the directly-downloaded cfitsio version because it doesn't have the patch.  What do we do in this case?

esheldon:
for bzip we have the `SKIP_BZIP_TEST` check.  If that environment variable is set, bzip isn't tested.  could do the same for HCOMPRESS

erykoff:
Sounds like a plan.

erykoff:
@esheldon All right!  Got the lib checking in, and all checks have passed.  I also checked that it works on MacOS (High Sierra) with compiled or external lib.

esheldon:
Have you done checks that everything works when curl is not detected?  Is there something we should add to the tests or travis for that case?

erykoff:
I've now checked that everything works when curl is not detected, but the only way I could figure out how to do this was manually hack the external cfitsio Makefile to remove curl support.  There's no way of getting cfitsio `configure` to turn off curl support as far as I can tell, so it would be hard to automate in travis.

esheldon:
one way would be for the travis base OS to not have the curl library installed.

But I'm satisfied.

PR #162, opened 2018-04-13
Removed initial numpy import

albireox:
In `setup.py` the numpy import is wrapped in a try and except. This allows, for instance, to install fitsio in a read the docs build where numpy has not yet been installed. There is, however, an `import numpy` at the top of the file that defeats this purpose. This PR gets rid of it.


albireox:
Any new on this being merged anytime soon?

esheldon:
I'm very sorry I lost track of this one.  thanks for the PR

PR #168, opened 2018-07-24
Updates to enable GZIP_2 compression

menanteau:
These changes enable GZIP_2 compression which is now available on newer version of cfitsio.


esheldon:
thanks I'll have a look soon

PR #171, opened 2018-10-04
FIX: eliminate sdist warnings

rainwoodman:
I tested after this PR the package can be uploaded to test.pypi.org via

twine upload --repository-url https://test.pypi.org/legacy/ dist/fitsio-1.0.0rc1.tar.gz


rainwoodman:
This PR addresses the issue raised at

https://github.com/esheldon/fitsio/pull/133#issuecomment-427040532

esheldon:
I dread a move to setuptools, it has caused me a lot of headaches in the past.  This will require a lot of testing for usability, in the real world, before I would consider a merge.

esheldon:
A good example is that this will now install into a directory like this
```
lib/python3.6/site-packages/fitsio-1.0.0rc1-py3.6-linux-x86_64.egg/fitsio/
```
and this will conflict with old installs, things won't work correctly.

rainwoodman:
Sorry but I had the same resentment towards setuptools until I realized a conversion to setuptools is inevitable. Most main stream packages have switched, including numpy:

https://github.com/numpy/numpy/blob/master/setup.py#L392 

If it worked out for numpy, it should work out for fitsio as well.

My understanding is that pip is "deprecating" support to raw distutils -- one of the arguments I saw, was that raw distutils packages cannot be 'properly uninstalled'. The change to the installation directory was a necessary evil to support proper version management in pip.

As a result, some growing pain has to be endured for this change, like the one you have. My method of coping was to manually purge the previous, distutils installation, if any conflicts rises. This seem to affect developers more than the end users. 

If you want setuptools to write to the old directory, I think you can pass in the flag --single-version-externally-managed. This is necessary in the conda recipe.

esheldon:
OK, let's take the dive.

Are you ready to merge or are you planning any further changes?

rainwoodman:
Great. I don't have more planned changes.

Issue #173, opened 2018-10-05
Ability to write undefined values (i.e. None) to headers


menanteau:
From Pence et al 2010, the standard for null/empty/undefined should be as follows: 
```
Note that there is a subtle distinction between the following three keywords:
KEYWORD1= '' / null string keyword
KEYWORD2= ' ' / empty string keyword
KEYWORD3=    / undefined keyword

The value of KEYWORD1 is a null, or zero length string whereas the value of the KEYWORD2 is an empty string (nominally a single space character because the first space in the string is significant, but trailing spaces are not). The value of KEYWORD3 is undefined and has an indeterminate data type as well, except in cases where the data type of the specified keyword is explicitly defined in this standard.
```
At the moment the behavior for None is not respected. 

```
fitsio.write('test.fits',zeros( (10,10)), clobber=True, header={'Noneval':None,'Nones':'None'})
h=fitsio.read_header('test.fits')
h
ONEVAL = '        '
NONES   = 'None    '
```

esheldon:
working on this in branch undefined

esheldon:
please try out the undefined branch

PR #174, opened 2018-10-05
Deal properly with undefined header keywords

esheldon:
This addresses issue #173 

when writing, a None value for a header keyword will trigger writing a header key with undefined (blank) value.

when reading, these are converted to None values


esheldon:
this failure seems to be related to wget timing out, see this issue: #175

esheldon:
I think the travis failures are just due to wget failing, so I will go ahead and merge.

PR #180, opened 2018-11-02
ENH numpy 'U' dtype support for python 3

### LINKED ISSUE #179: support unicode strings
I am making this issue to start a convo around unicode string support for python 3. TBH, this looks like a nightmare. 

Some notes:
- I checked the `cfitsio` docs and it does not appear that they have support for unicode. 
- Right now `fitsio` doesn't itself support the 'U' numpy type (which denotes unicode strings). 

So to make this happen, I think we have to put in manual conversions. On writing, this is straight forward since we would know it is a unicode type before hand and so could convert to a string of bytes (assuming we stick to a fixed length codec, which might cause overhead if we have to convert). 

Reading is the real nightmare though. Basically, if we see a column of bytes, IDK how we would know it was written from a column of unicode strings in python. I was thinking we could reserve `TSTRING` in the fits standard for `NPY_UNICODE` objects, but then if someone wants to write an array of python bytes data from `NPY_STRING`, we would have to use the `TBYTE` fits type. However, this would be confused with an array of unsigned one-byte integers, `NPY_UINT8`. 

Thus we have a many to one issue in that the FITS types are not expressive enough. 

This leaves us with truly awful solutions around monkey patching the header info to hold the extra information on which columns need unicode conversion on read. :donotwant:

So, unless, I am mistaken, we would have to basically write unicode into the FITS standard before we can support it in python 3. 

Thoughts @rainwoodman @esheldon @erykoff ?

beckermr:
This PR has support for numpy 'U' dtypes in python 3.

Closes #179

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,12 @@ env:
     - PYTHON_VERSION=3.5 NUMPY_VERSION=1.9
     - PYTHON_VERSION=3.5 NUMPY_VERSION=1.11
 
+    - PYTHON_VERSION=3.6 NUMPY_VERSION=1.9
+    - PYTHON_VERSION=3.6 NUMPY_VERSION=1.15
+
+    - PYTHON_VERSION=3.7 NUMPY_VERSION=1.9
+    - PYTHON_VERSION=3.7 NUMPY_VERSION=1.15
+
 before_install:
     - if  [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       wget http://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -O miniconda.sh;
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -29,11 +29,14 @@ import warnings
 from . import _fitsio_wrap
 from .util import FITSRuntimeWarning, cfitsio_version
 
+if sys.version_info >= (3, 0, 0):
+    IS_PY3 = True
+else:
+    IS_PY3 = False
+
 # for python3 compat
-try:
-    xrange=xrange
-except:
-    xrange=range
+if IS_PY3:
+    xrange = range
 
 from functools import reduce
 
@@ -1720,6 +1723,11 @@ class TableHDU(HDUBase):
             # some logic
             data_send = array_to_native(data, inplace=False)
 
+        if IS_PY3 and data_send.dtype.char == 'U':
+            # for python3, we convert unicode to ascii
+            # this will error if the character is not in ascii
+            data_send = data_send.astype('S', copy=False)
+
         self._FITS.write_column(self._ext+1, colnum+1, data_send, 
                                 firstrow=firstrow+1, write_bitcols=self.write_bitcols)
         del data_send
@@ -1773,7 +1781,11 @@ class TableHDU(HDUBase):
         if name in self._colnames:
             raise ValueError("column '%s' already exists" % name)
 
-        descr=data.dtype.descr
+        if IS_PY3 and data.dtype.char == 'U':
+            descr = data.astype('S', copy=False).dtype.descr
+        else:
+            descr = data.dtype.descr
+
         if len(descr) > 1:
             raise ValueError("you can only insert a single column, "
                              "requested: %s" % descr)
@@ -2014,6 +2026,8 @@ class TableHDU(HDUBase):
 
             self._FITS.read_as_rec(self._ext+1, 1, nrows, array)
 
+            array = self._maybe_decode_fits_ascii_strings_to_unicode_py3(array)
+            
             for colnum,name in enumerate(array.dtype.names):
                 self._rescale_and_convert_field_inplace(array,
                                           name,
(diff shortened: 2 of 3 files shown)


esheldon:
Thanks Matt this is awesome. I'll have a look

.travis.yml
```diff
@@ -32,6 +32,12 @@ env:
     - PYTHON_VERSION=3.5 NUMPY_VERSION=1.9
     - PYTHON_VERSION=3.5 NUMPY_VERSION=1.11
 
+    - PYTHON_VERSION=3.6 NUMPY_VERSION=1.9
+    - PYTHON_VERSION=3.6 NUMPY_VERSION=1.15
+
+    - PYTHON_VERSION=3.7 NUMPY_VERSION=1.9
+    - PYTHON_VERSION=3.7 NUMPY_VERSION=1.15
+
```
beckermr: We may want to dump the older python builds.

fitsio/fitslib.py
```diff
@@ -29,11 +29,14 @@
 from . import _fitsio_wrap
 from .util import FITSRuntimeWarning, cfitsio_version
 
+if sys.version_info >= (3, 0, 0):
```
beckermr: This was `sys.version_info > (3, 0, 0)` in another place in the code. IDK why. Just noting it here.

fitsio/fitslib.py
```diff
@@ -1720,6 +1723,11 @@ def write_column(self, column, data, **keys):
             # some logic
             data_send = array_to_native(data, inplace=False)
 
+        if IS_PY3 and data_send.dtype.char == 'U':
+            # for python3, we convert unicode to ascii
+            # this will error if the character is not in ascii
+            data_send = data_send.astype('S', copy=False)
```
beckermr: This block should not be needed since this was done in `array_to_native`. However, for whatever reason, the tests would not pass without it. IDK.

fitsio/fitslib.py
```diff
@@ -1773,7 +1781,11 @@ def insert_column(self, name, data, colnum=None):
         if name in self._colnames:
             raise ValueError("column '%s' already exists" % name)
 
-        descr=data.dtype.descr
+        if IS_PY3 and data.dtype.char == 'U':
+            descr = data.astype('S', copy=False).dtype.descr
```
beckermr: It might be better to adjust the descriptor here instead of casting the whole array.

fitsio/fitslib.py
```diff
@@ -2505,17 +2532,22 @@ def _read_rec_with_var(self, colnums, rows, dtype, offsets, isvar, vstorage):
                     # storing in object array
                     # get references to each, no copy made
                     for irow,item in enumerate(dlist):
+                        if IS_PY3 and isinstance(item, bytes):
+                            item = item.decode('ascii')
```
beckermr: I'm not so sure about testing each object in an array. This might muck with other things or cause random decoding errors on things not meant to be decoded. In general, we cannot know if a bytes object is supposed to be a string or something else.

fitsio/fitslib.py
```diff
@@ -2696,12 +2750,16 @@ def _get_tbl_numpy_dtype(self, colnum, include_endianness=True):
                 addstr=''
             else:
                 addstr='>'
-            if npy_type not in ['u1','i1','S']:
+            if npy_type not in ['u1','i1','S','U']:
```
beckermr: Looking at it now, I am wondering about this change.

fitsio/fitslib.py
```diff
@@ -3082,9 +3146,13 @@ def read(self, **keys):
                     # storing in object array
                     # get references to each, no copy made
                     for irow,item in enumerate(dlist):
+                        if IS_PY3 and isinstance(item, bytes):
+                            item = item.decode('ascii')
```
beckermr: Again not sure about decoding generic objects.

beckermr:
Bump on this @esheldon. I'd love to merge this PR this week if possible.

esheldon:
I haven't been able to test this in the real world yet, nor has anyone else but you.  It is a very big (and important) change that I think need testing in the wild.

beckermr:
Sounds good. I'll plug away on it and see what rustles up!

esheldon:
I made a branch unicode, can you please do the PR against that?  That will make it easy to test

esheldon:
OK, merged this.  We can do a new PR into master after I do some tests

beckermr:
We should look at my comments above and maybe address them BTW.

.travis.yml
beckermr: We may need to move away from travis anyhow. Resolving this comment.

fitsio/fitslib.py
beckermr: after discussion with @esheldon I think we are good on this. The only thing in the C lib that comes back out as bytes are TSTRING items and those should always be strings.

fitsio/fitslib.py
beckermr: Looking again, we never read unicode bytes directly. The default numpy unicode type is utf32 IIUIC and this can be stored in big or little endian. Straight bytes cannot be. I don't think want to force the unicode representation in the numpy dtype. Resolving this.

fitsio/fitslib.py
beckermr: I fixed this. Had to do with an edge case for in place array conversion.

PR #186, opened 2018-11-29
Nullstr

esheldon:
make cfitsio obey standards with regards to strings in binary tables

to do:
- [ ] move changes to diff file of the cfitsio code that gets applied on the fly
- [ ] make sure we write tests in cfitsio of null handling for strings


cfitsio3430patch/fitsio.h
```diff
@@ -797,6 +797,7 @@ int CFITS_API fits_delete_iraf_file(const char *filename, int *status);
 /*---------------- utility routines -------------*/
 
 float CFITS_API ffvers(float *version);
+int CFITS_API fits_use_standard_strings(void);
```
beckermr: So here, we prob want to make a short name for this function and then put the long name in longname.h.

esheldon: Maybe, but note there is precedence for long name only.  See the compression routines toward the bottom

esheldon: Also, the short ones are legacy from fortran, and our new feature would not be applicable in fortran where there are no null terminated strings.

beckermr: Ahhhhh cool. Then let's punt.

beckermr:
Test failure appears to be related to distutils/setuptools not fully cleaning up the compiled extension. Might be good to `rm -rf` all of the egg and build dirs in travis.

beckermr:
Some notes (mostly for me):
 - this [line](https://github.com/esheldon/fitsio/blob/master/cfitsio3430patch/buffers.c#L134) is ok - it fills zero (which is a null character for strings, and correctly fills 32 (which is a space) for fits ascii tables.
- I cannot tell if this [line](https://github.com/esheldon/fitsio/blob/master/cfitsio3430patch/putcolu.c#L165) needs to be changed. There is another block below it that is similar that I am also confused about.

fitsio/fitsio_pywrap.c
```diff
@@ -4131,6 +4131,16 @@ PyFITS_cfitsio_version(void) {
...
 }
 
+static PyObject *
+PyFITS_cfitsio_use_standard_strings(void) {
+    if ( fits_use_standard_strings() ) {
+        Py_RETURN_TRUE;
+    } else {
+        Py_RETURN_FALSE;
+    }
+}
```
beckermr: We're going to have to figure out what to do if an older version of cfitsio doesn't have this function when linking. We might need a further compile-time check w/ a macro when building the wrapper.

beckermr: This is why the tests are failing.

setup.py
```diff
@@ -1,7 +1,6 @@
 from __future__ import print_function
-import distutils
-from distutils.core import setup, Extension, Command
-from distutils.command.build_ext import build_ext
+from setuptools import setup, Extension, Command
+from setuptools.command.build_ext import build_ext
```
beckermr: I know you hate setuptools. Sorry. This is not needed for the code to work, but it got rid of a warning about an ignored option (`install_requires`). This option makes the code play nicer with `pip` and so it might be time to switch. :(

esheldon: this is like when a bill gets written that everyone wants, but then one of the parties writes in something extra that people don't want because they think the bill will get passed no matter what :)

esheldon:
thanks for getting this to work with system cfitsio

Next is to add more tests

esheldon:
I think the `testTableResize` test already addresses the issue of expanding tables and filling with default values.

I'm somewhat surprised that it works, because I thought the code in `putcolu` would fill with blanks.

beckermr:
I cannot figure out where the code in putcolu gets used.

beckermr:
As a followup here, it appears even the cfitsio test program doesn't call the null str function. I have said before that one should stop testing when you stop feeling fear, shame and guilt. I am feeling a lot of guilt about not resolving this, which means we may want to add a test to cfitsio too. :(

esheldon:
should we go ahead and merge this into unicode and open new issues for the unresolved stuff?

beckermr:
Ugh Maybe? 

We really should pull the patches to vendored source code out so we can push them up upstream. IDK if that is easier now or later...

Also those patches probably don't catch everything in the C library since the python wrapper doesn't touch all code paths.

esheldon:
yeah, that's what I meant by "unresolved stuff" :)

beckermr:
yep. ugh no good answers. yolo. let's merge away! 🤞

beckermr:
probably good to make a merge commit to keep the diffs as opposed to squashing

esheldon:
right this is only merging into the unicode branch, so not a huge commitment

PR #194, opened 2018-12-14
ENH add support for python 3 strings and proper null termination

### LINKED ISSUE #179: support unicode strings

### LINKED ISSUE #185: make fitsio strings match the standard
This issue has been brought up before.

cfitsio does not match the standard with regard to strings.  It converts null characters in the input strings to a space and always pads with spaces to the declared column size.  This is because cfitsio predates the standard and they want to maintain backward compatibility.

This padding with spaces is pervasive.  e.g. cfitsio will expand a table to have new rows if a single column in the table is extended, and the string columns are filled with spaces to the declared width.  So it is not a complete solution to just implement up our own string writing outside of cfitsio, because such operations as above will still result in spaces everywhere.

Probably the best thing to do is patch the sources.  an attempt was made in the nullstr branch that was not successful.

beckermr:
This PR contains a variety of updates for v1 including

- support for python 3 strings
- support in cfitsio for null terminated strings
- improved build system with setuptools and patches
- various bug fixes

Closes #179 
Closes #185


fitsio/fitsio_pywrap.c
```diff
@@ -40,6 +40,13 @@ struct PyFITSObject {
     fitsfile* fits;
 };
 
+#ifdef FITSIO_PYWRAP_ALWAYS_NONSTANDARD_STRINGS
+static int fits_use_standard_strings(void)
+{
+    return 0;
+}
+#endif
+
```
beckermr: @esheldon Why don't we want this macro around only the return value?

beckermr: Ohhhhh. This function is here in case the version of fitsio doesn't have this symbol. NM.

esheldon: I think that's correct

beckermr:
OK. I round tripped all of the data from python 2, python 3, astropy, and both versions of fitsio. I got all of the things I expected. See the gist with the scripts here: https://gist.github.com/beckermr/897141e0a44ae2a192feebbfd42b40fa

beckermr:
This is ready to merge on my end @esheldon.

Issue #207, opened 2019-04-05
Header parsing wrongly tries to cast strings to floats


dstndstn:
We encountered this one in the wild.  fitsio tries to convert a string like '89113e6' into a float!

Minimal test case:

```
import fitsio
hdr = fitsio.FITSHDR()
val = '89113e6'
hdr['PLPROCID'] = val
fitsio.write('/tmp/1.fits', None, header=hdr)
hdr2 = fitsio.read_header('/tmp/1.fits')
val2 = hdr2['PLPROCID']
print(val2)
assert(val2.strip() == val.strip())
```

Prints `89113000000.0`

It seems like the parsing rule ought to be that if a value is in quotes, it's a string.

esheldon:
Hi @dstndstn please check out the fix in PR #208

PR #208, opened 2019-04-05
#207 fix issues converting string header entries

esheldon:
If the entry is a string, don't even try to convert it.

Also remove extra convert attempt; this is not needed because conversion happens in the FITSCard

### DIFF
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -1465,7 +1465,7 @@ class HDUBase(object):
         number.
         """
         # note converting strings
-        return FITSHDR(self.read_header_list(), convert=True)
+        return FITSHDR(self.read_header_list())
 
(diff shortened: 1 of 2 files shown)


esheldon:
@dstndstn the CI has passed, will you have a chance to test it out or should I just merge it?

dstndstn:
Hi @esheldon , Thanks for the amazingly fast turn-around!  @moustakas (who diagnosed this) is checkout out this solution, so if you can hold on an hour, can confirm the fix.  Thanks again!

esheldon:
Of course, no hurry from my end.

moustakas:
Just tested and works like a charm, thank you @esheldon!

esheldon:
you're welcome!

PR #211, opened 2019-04-08
update changelog and readme

### LINKED ISSUE #205: add README docs for new unicode features and cut tag v1.0.0rc2

beckermr:
Closes #205

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,13 @@ New Features
 
     - Support for python 3 strings.
     - Support for proper string null termination.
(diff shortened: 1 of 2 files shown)


beckermr:
This is the last PR we need before tagging v1.0.0 @esheldon

beckermr:
Note that I did not actually bump the version to 1.0.0. I can do that here as well.

esheldon:
Let's hold off on actually putting v1.0.0 in there, so as to avoid messing up codes that store that string for reference

PR #212, opened 2019-04-09
raise ValueError if strings have size 0. Add test

### LINKED ISSUE #210: make sure to error on zero sized column types (e.g., 'U0', 'S0')

esheldon:
closes #210

### DIFF
--- a/fitsio/test.py
+++ b/fitsio/test.py
@@ -1316,6 +1316,10 @@ class TestReadWrite(unittest.TestCase):
 
                     except ValueError:
                         got_error=True
+                    except TypeError:
+                        # old numpy don't allow this dtype, so will throw
+                        # a TypeError for empty dtype
+                        got_error=True
 
                     self.assertTrue(got_error == True,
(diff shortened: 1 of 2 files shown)


fitsio/test.py
```diff
@@ -1316,6 +1316,10 @@ def testTableWriteBadString(self):
 
                     except ValueError:
                         got_error=True
+                    except TypeError:
```
beckermr: Can we scope this error to only coming from numpy?

esheldon: It doesn't look to be subclassed.  do you have something else in mind?

beckermr: Maybe test against the exact error string from numpy?

beckermr: Or a substring of it?

```python
except TypeError as e:
    if str(e) == 'Empty data-type':
         got_error = True
    else:
        raise
```

beckermr: Not sure if raise is right for the else clause

Issue #176, opened 2018-10-05
Comment is lost when modifying a KEYWORD in a FITSHDR object


menanteau:
There seems to be a bug in the FITSHDR such that when the value of a KEYWORD in header is modified, when written, the comment part for that KEYWORD is lost.
For example:

```
l1 = 'KEY1    =                   77 / My comment1'
l2 = 'KEY2    =                   88 / My comment2'
hdr=fitsio.FITSHDR()
hdr.add_record(l1)
hdr.add_record(l2)
hdr
KEY1    =                   77 / My comment1
KEY2    =                   88 / My comment2
```

But it we modify the KEYWORD for `KEY1`
```
hdr['KEY1'] = 99
hdr
KEY1    =                   99
KEY2    =                   88 / My comment2
```

We lose the comment for `KEY1`

esheldon:
closed via #213

PR #213, opened 2019-04-09
ensure comment is preserved when setting header value of existing entry

esheldon:
add a test

addresses #176


esheldon:
@menanteau this PR addresses your issue #176

PR #217, opened 2019-04-11
Fix long string reading in FITS headers

esheldon:
long strings are read correctly (via CONTINUE)

now all keyword values are read into long strings and converted in the python code.

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -3972,17 +3972,24 @@ PyFITSObject_read_header(struct PyFITSObject* self, PyObject* args) {
     int status=0;
     int hdunum=0;
     int hdutype=0;
+    char *tmp=NULL;
+    int lcont=0, lcomm=0, ls=0;
+    int tocomp=0;
+    char *longstr=NULL;
 
     char keyname[FLEN_KEYWORD];
     char value[FLEN_VALUE];
     char comment[FLEN_COMMENT];
     char card[FLEN_CARD];
+    long is_string_value=0;
 
     int nkeys=0, morekeys=0, i=0;
 
     PyObject* list=NULL;
     PyObject* dict=NULL;  // to hold the dict for each record
 
+    lcont=strlen("CONTINUE");
+    lcomm=strlen("COMMENT");
 
     if (!PyArg_ParseTuple(args, (char*)"i", &hdunum)) {
         return NULL;
@@ -4002,12 +4009,11 @@ PyFITSObject_read_header(struct PyFITSObject* self, PyObject* args) {
         return NULL;
     }
 
-    list=PyList_New(nkeys);
+    list=PyList_New(0);
     for (i=0; i<nkeys; i++) {
 
         // the full card
         if (fits_read_record(self->fits, i+1, card, &status)) {
-            // is this enough?
             Py_XDECREF(list);
             set_ioerr_string_from_status(status);
             return NULL;
@@ -4016,21 +4022,46 @@ PyFITSObject_read_header(struct PyFITSObject* self, PyObject* args) {
         // this just returns the character string stored in the header; we
         // can eval in python
         if (fits_read_keyn(self->fits, i+1, keyname, value, comment, &status)) {
-            // is this enough?
+            Py_XDECREF(list);
+            set_ioerr_string_from_status(status);
+            return NULL;
+        }
+        ls=strlen(keyname);
+        tocomp = (ls < lcont) ? ls : lcont;
+        if (strncmp(keyname,"CONTINUE",tocomp)==0) {
+            continue;
+        }
+
+        if (fits_read_key_longstr(self->fits, keyname, &longstr, comment, &status)) {
             Py_XDECREF(list);
             set_ioerr_string_from_status(status);
             return NULL;
         }
 
-        dict = PyDict_New();
-        add_string_to_dict(dict,"card_string",card);
-        add_string_to_dict(dict,"name",keyname);
-        add_string_to_dict(dict,"value",value);
-        add_string_to_dict(dict,"comment",comment);
+        if (card[8]=='=' && card[10]=='\'') {
+            is_string_value=1;
+        } else {
+            is_string_value=0;
+        }
 
-        // PyList_SetItem and PyTuple_SetItem only exceptions, don't 
-        // have to decref the object set
-        PyList_SetItem(list, i, dict);
+
+        dict = PyDict_New();
+        add_string_to_dict(dict,"name",keyname);
+        add_string_to_dict(dict,"comment",comment);
+        add_long_to_dict(dict, "is_string_value", is_string_value);
+
+        // if not a comment but empty value, put in None
+        tocomp = (ls < lcomm) ? ls : lcomm;
+        if (0==strlen(longstr) && strncmp(keyname,"COMMENT",tocomp)!=0) {
+            add_none_to_dict(dict, "value");
+        } else {
+            add_string_to_dict(dict,"value",longstr);
+        }
+
+        free(longstr);
+
+        PyList_Append(list, dict);
+        Py_XDECREF(dict);
 
     }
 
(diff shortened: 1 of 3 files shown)


esheldon:
@kadrlica this seems to do the job

fitsio/fitslib.py
```diff
@@ -4487,7 +4488,8 @@ def _record2card(self, record):
         v_isstring=isstring(value)
 
         if name == 'COMMENT':
-            card = 'COMMENT   %s' % value
+            #card = 'COMMENT   %s' % value
+            card = 'COMMENT %s' % value
```
beckermr: Did this need two spaces between `COMMENT` and `%s`?

esheldon:
This addresses #117 #118

fitsio/fitslib.py
esheldon: the old way added extra spaces; this may have to do with the new way that things are being read

beckermr: Do we need to change the code block with CONTINUE below as well then?

esheldon: I don't know.  That code path is not hit by the tests, I do not think because the CONTINUE are no longer in the list returned from the C code

esheldon: note the only place record2card is used is when you print the header, and that is not intended for use in generating a valid header.

Of course it would be good to get it right

esheldon: OK, just ran a test that shows continues are being printed correctly there

esheldon:
I added a test of creating a header from cards, including a continue

fitsio/fitsio_pywrap.c
```diff
@@ -4016,21 +4022,46 @@ PyFITSObject_read_header(struct PyFITSObject* self, PyObject* args) {
...
+
+        // if not a comment but empty value, put in None
+        tocomp = (ls < lcomm) ? ls : lcomm;
+        if (0==strlen(longstr) && strncmp(keyname,"COMMENT",tocomp)!=0) {
+            add_none_to_dict(dict, "value");
+        } else {
+            add_string_to_dict(dict,"value",longstr);
+        }
+
+        free(longstr);
```
esheldon: I should set `longstr=NULL` after the free

beckermr: I didn't look at the C. :p

beckermr:
I'm not good enough at python reference counting to make coherent comments on the C. Sorry. :(

kadrlica:
I'll try to have a look at the C code (though I don't claim much expertise).

esheldon:
I'm pretty sure the ref counting is fine here, except things get a bit unpredictable for me when errors occur in the python operations.  But those should be extremely rare and would signal a fatal error anyway.

beckermr:
Alright @esheldon. If you are happy with the tests, then we should simply merge.

esheldon:
This new code did not deal with empty string values correctly, turned out we didn't have a unit test for that.

kadrlica:
What about the corrupted continue cards? These occur in a non-zero fraction on DES images.

esheldon:
Do you have a test for that?

esheldon:
I see it won't work, need to check for the `=` sign in the C code, but it would be best if you can give me a very simple test

esheldon:
@kadrlica I came up with something, it is in the tests now. Please see if this will cover your cases

kadrlica:
I'm looking back at #130 and I think your re-implementation of those old tests works. There was a third case (described [here](https://github.com/esheldon/fitsio/issues/130#issuecomment-317634351)) that we were dealing with:
```
CONTINUE = value / comment
```
but I'm not sure if this was motivated by a real example of a corrupted header like this. I can go try to parse 100k DES headers with your new branch and see if I hit any problems.

esheldon:
currently I'm taking the view "it doesn't crash so we are ok", i'm not trying to make sure we can make sense of the output

esheldon:
in other words, people should not expect to recover information from those CONTINUE=

beckermr:
I agree w/ @esheldon. If the fits file is corrupt, we cannot be expected to produce sensible output. We should however not crash unless the corruption is egregious that it is no longer possible to go on.

esheldon:
Just to be clear, the other header entries should be fine, just these continue are probably not going to be easy to extract information from

esheldon:
In fact I punted and these are not returned as entries; if you want different behavior we could do something but I don't how useful or interesting it would be

kadrlica:
Ok, I've run a test over a random set of 100k DES images that were processed with the old version of `fitsio`. I've confirmed that some fraction of these images had corrupted `CONTINUE` headers due to processing with the old version of fitsio (mostly in the `PROGRAM` header keyword).

The new version from this PR is able to load the headers of all these images without crashing, and appears to successfully load the first card of the long keywords, but ignores subsequent lines and includes corrupted information for the comment strings.

```
PROGRAM = Search for ultra-faint Milky Way Satellite Galaxies in the & / pU??K
```

I agree with the sentiments above that this is a suitable solution (I'll note that we do have a bit more responsibility to read these headers since it was fitsio that corrupted them in the first place).

fitsio/fitslib.py
```diff
@@ -4607,9 +4627,35 @@ def verify(self):
...
+            else:
+                value = ast.literal_eval(value_orig)
+        except:
+            value = self._convert_string(value_orig)
+
+        if isinstance(value,int) and '_' in value_orig:
+            value = value_orig
+
+        return value
+
```
beckermr: We removed this code because of another bug. I guess we have a test for that bug already so this is fine?

esheldon: this is only used for the specific case we have here, it should be fine.

esheldon:
As for responsibility, I'm happy to take code to do something more.

For the vestigal stuff, please do let me know if you see a problem spot.

beckermr:
@kadrlica on your test on the 100k images above, it only ignores subsequent stuff when things are corrupted, right?

esheldon:
So it is a small change to go ahead and return the continue= stuff, but it ends up looking quite odd.  e.g. for the case I used the value was empty and the comment had a bizarre combination of stuff in it

kadrlica:
I've gone and checked ~10k raw DECam files that should have "good" headers. The CONTINUE cards appear to be parsed, but they have a strange trailing '&'. The comment also appears to have been duplicated. The extra '&' does seem to be there in the card, but maybe an older formatting?

Here's an example of the native cards
```
PROGRAM = 'Search for Ultra-Faint Milky Way Satellite Galaxies and Star &'      CONTINUE  'Clusters in the Southern Hemisphere&'                                CONTINUE  '&' / Current observing orogram 
```
And here is how it is parsed by fitsio:
```
{'comment': ' Current observing orogram Current observing orogram', 'is_string_value': 1L, 'name': 'PROGRAM', 'value': 'Setting the Scale: Determining the Absolute Mass Normalization and Scaling Relations for Clusters at z~0.1&'}
```
astropy.io.fits appears to do fine with this:
```
'Setting the Scale: Determining the Absolute Mass Normalization and Scaling Relations for Clusters at z~0.1'
```

kadrlica:
I've added a simple example of this type of header in `testCorruptContinue` (this isn't a corrupted continue, but I wasn't sure where/if you'd want to keep this).

esheldon:
That one is corrupt in an odd way.

If I write out that string the header looks like this
```
PROGRAM = 'Setting the Scale: Determining the Absolute Mass Normalization and &'
CONTINUE  'Scaling Relations for Clusters at z~0.1' / Current observing orogram
```
And it gets read in correctly.

esheldon:
You actually mixed two examples above, one about milky way satellites one about galaxy clusters, so its a bit hard to interpolate

kadrlica:
Sorry, looking at two separate examples of the same issue (this is very common among raw DECam exposures). I've updated the example so they are consistent. If you say this is corrupt than fine, but this was not caused by fitisio, its read by astropy.

esheldon:
It is certainly commendable that astropy reads it.  I would like to make it work, but I'm feeling the need to move on for the moment, and cut v1.0

kadrlica:
Yeah, I understand, I was just unsure that this was really a corruption.

esheldon:
To be clear: reading it would require a fix in cfitsio, which is a different sort of matter.

kadrlica:
I think this is the consistent format used by DECam, so that's just something to be aware of.

esheldon:
Are you saying that is the original  header, not processed in any way?  Can you please point me to an example (offline is ok)

kadrlica:
Yes, that's what I'm saying, and I'll take it offline.

PR #221, opened 2019-04-14
Add deprecation warning indexing by scalar.

### LINKED ISSUE #177: Indexing a scalar item from a TABLE_HDU's column returns a vector.
Assuming f[1] is a TABLE_HDU with a single DEC column.
```
with .... as f:
    assert f[1]['DEC'][:].ndim = 1
    assert f[1]['DEC'][0].ndim = 0 # fails, the value is one
```

This is version 0.9.11. Has this been brought up before?

The usual indexing convention is to return a scalar in this case, for example if you are indexing from h5py or a numpy array.

rainwoodman:
Also mention the behavior will change to align with numpy / h5py.

Fix #177

### DIFF
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -3007,10 +3007,18 @@ class TableHDU(HDUBase):
 
             fits=fitsio.FITS(filename)
             fits[ext][:]
+            fits[ext][2]   # see below (Indxing with a scalar)
             fits[ext][2:5]
             fits[ext][200:235:2]
             fits[ext][rows]
             fits[ext][cols][rows]
+
+        Indexing with a scalar:
+
+            Before version 1.1 returns fits[ext][2:3].
+            And a DeprecationWarning is issued.
+            Since version 1.1 returns a numpy scalar (as indexing a
+            ndarray).
 
         Note data are only read once the rows are specified.
 
@@ -3692,8 +3700,18 @@ class TableColumnSubset(object):
             self.fitshdu._process_args_as_rows_or_columns(arg, unpack=True)
         if isrows:
             # rows was entered: read all current column subset
-            return self.read(rows=res)
-
+            r = self.read(rows=res)
+            if numpy.isscalar(res):
+                mess=("Indexing a TableColumnSubset with "
+                      "a scalar currently returns a length 1 array. "
+                      "The behavior will change to return a scalar in the "
+                      "next release, consistent with h5py and numpy. "
+                     )
+                # FIXME: enable this and change the test case
+                # testTableColumnIndexScalar.
+                warnings.warn(mess, DeprecationWarning)
+                # r = r[0]
+            return r
         # columns was entered.  Return a subset objects
         return TableColumnSubset(self.fitshdu, columns=res)
 
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -2858,32 +2858,42 @@ class TableHDU(HDUBase):
         Can also be tuples or arrays.
         """
 
-        isslice = False
-        isrows = False
-        result=arg
-        if isinstance(arg, (tuple,list,numpy.ndarray)):
+        flags = set()
+        #
+        if isinstance(arg, (tuple, list, numpy.ndarray)):
             # a sequence was entered
             if isstring(arg[0]):
-                pass
-            else:
-                isrows=True
                 result = arg
+            else:
+                result = arg
+                flags.add('isrows')
         elif isstring(arg):
             # a single string was entered
-            pass
+            result = arg
         elif isinstance(arg, slice):
-            isrows=True
             if unpack:
+                flags.add('isrows')
                 result = self._slice2rows(arg.start, arg.stop, arg.step)
             else:
-                isslice=True
+                flags.add('isrows')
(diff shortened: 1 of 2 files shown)


beckermr:
We cannot push this change on 1.1 since it is api breaking. We either need to retag 1.0 which might still be possible or we have to wait for 2.0.0

rainwoodman:
A major release 2.0.0 might be too far away. I highly doubt many people actually rely on the scalar index in production code, and it is an easy fix in that case -- thus something like a 1.0.1 that throws a warning, and wait for the next minor release could be good enough.  I shall change this to a FutureDeprecationWarning, as well.

beckermr:
As a statement of fact, we actually have *no idea* if people are relying on this or not. This is why it makes sense to be pretty strict on versioning.

@esheldon has the final say ofc.

esheldon:
The main way people get fitsio is through conda and pypi.  I think @beckermr set it up conda so it only gets triggered if we push to pypi, and we haven't pushed ot pypi.  So I'd be happy to tag 1.0.1 with the numpy-style scalar index and then make the first 1.* release be 1.0.1, and then push to pypi.

Does that make sense?

beckermr:
This will potentially break code for anyone who is pulling directly from github, but I'd be ok with this because we have not widely released the source yet. We should leave the 1.0.0 tag in place though.

beckermr:
If we do this, I suggest we pass on the warning and just release the new code.

esheldon:
Agreed.

I just tried this code and I didn't get a warning anyway, so I'm a bit confused.

beckermr:
Well that's weird...

fitsio/fitslib.py
```diff
@@ -3692,8 +3700,18 @@ def __getitem__(self, arg):
             self.fitshdu._process_args_as_rows_or_columns(arg, unpack=True)
         if isrows:
             # rows was entered: read all current column subset
-            return self.read(rows=res)
-
+            r = self.read(rows=res)
+            if numpy.isscalar(res):
```
beckermr: This code looks wrong anyways. See the documentation of numpy.isscalar, https://docs.scipy.org/doc/numpy/reference/generated/numpy.isscalar.html

esheldon: Well, it is detecting the input is a scalar but the warning doesn't happen.

esheldon:
I tried commenting out the `r=r[0]` and it does return a scalar for the case like this
```python
f=fitsio.FITS(fname)
f[ext][colname][5]
```
However, it does not work for returning an entire row
```python
f[ext][5]
```
The numpy behavoir is to also return a scalar in that case.  Still not sure why the warning does not fire.

fitsio/fitslib.py
beckermr: The test here needs to be on how the user has indexed the array. In other words, we need to know if they passed something like '0:1` or `0`. The output is always a vector right now and so this if condition will always evaluate to `False`.

beckermr: We cannot test for a length 1 item since this would not properly distinguish the two cases either.

esheldon: that test is on `res` which is the rows, not the output `r`.

I agree we might want some more sophisticated handling here.

beckermr: Ack. OK. This PR needs work.

beckermr: Do we need a test for a tuple to catch whole rows?

beckermr: Wait no that won't work.

rainwoodman:
Some versions of python hides deprecation warning by default.
The behavior was changed to show it relatively recently.

rainwoodman:
I disagree. For the test case it falls to the branch where result is a
scalar.

I feel a proper fix calls for a refactor of the index parsing routine, but
can be done with some minimal impact patching there as well. I can take a
look in the evening if no one get to it sooner.

esheldon:
It did not return a scalar when I tried to get the entire row, only if I selected a particular field

rainwoodman:
Now it captures both cases. I missed one `__getitem__`. I thought Table was a subclass of `TableColumnSubset` -- wrong assumption.

fitsio/fitslib.py
rainwoodman: Changed to use ndim instead. I didn't know this!

esheldon:
Although 1.0 is released, it is not yet distributed because it has not been released through pypi or conda forge.  So I think we can make the first distributed release return scalars, and drop the warning.

esheldon:
I'm gonna merge, but I'll go in and remove the warning.  I'll update the CHANGES.md to list this and some other things as backwards incompatible changes

fitsio/fitslib.py
```diff
@@ -3007,11 +3017,19 @@ def __getitem__(self, arg):
...
             fits[ext][rows]
             fits[ext][cols][rows]
 
+        Indexing with a scalar:
+
+            Before version 1.1 returns fits[ext][2:3].
+            And a DeprecationWarning is issued.
+            Since version 1.1 returns a numpy scalar (as indexing a
+            ndarray).
+
```
beckermr: @esheldon This doc string is wrong.

PR #223, opened 2019-04-16
#102

### LINKED ISSUE #102: Speed (compared to pyfits and dfits/fitsort)
Hi,

I often need to read single header keywords from thousands of FITS files.

What I did in the beginning was using pyfits and read every single header and fetch the information that I need. So something like this:

```
from astropy.io import fits
hdr = [fits.getheader(f) for f in files]
```

I just did a test, and with 7000 FITS files this takes about 70 seconds.

So what I'm doing at the moment is a system call to dfits/fitsort, i.e. something like this:

```
dfits = subprocess.Popen(['dfits'] + files, stdout=subprocess.PIPE)
fitsort = subprocess.check_output(['fitsort'] + keywords, stdin=dfits.stdout).decode('utf-8')
dfits.wait()
```

This way, it only takes about 7 seconds to get the data from all files. But, as a downside, I need to now the keywords that I need before running this command (and I also need to parse the output).

Now I' m trying the same with fitsio, hoping for some improvement in speed as compared to pyfits. Getting close to the dfits/fitsort version would be nice. So this is what I do:

```
import fitsio
hdr = [fitsio.read_header(s) for s in args.spectra]
```

Unfortunately this also takes about 50 seconds, so not nearly as fast as I was hoping for.

Is there any other way to get this being done faster? Thanks.

Cheers,
Tim

esheldon:
closes #102

### DIFF
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -4272,7 +4272,8 @@ class FITSHDR(object):
             If the input is a card string, convert is implied True
         """
         if isinstance(record_in,dict) and 'name' in record_in and 'value' in record_in:
-            record = record_in
+            record = {}
+            record.update(record_in)
         else:
             record = FITSRecord(record_in)
 
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -116,9 +116,6 @@ def read_header(filename, ext=0, extver=None, case_sensitive=False, **keys):
     The FITSHDR allows access to the values and comments by name and
     number.
 
-    Under the hood, a FITS object is constructed and data are read using
-    an associated FITSHDU object.
-
     parameters
     ----------
     filename: string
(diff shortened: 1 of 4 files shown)


fitsio/fitslib.py
```diff
@@ -135,9 +135,37 @@ def read_header(filename, ext=0, extver=None, case_sensitive=False, **keys):
...
-    item=_make_item(ext,extver=extver)
-    with FITS(filename, case_sensitive=case_sensitive) as fits:
-        return fits[item].read_header()
+
+    dont_create=0
+    _fits =  _fitsio_wrap.FITS(filename, READONLY, dont_create)
+    try:
+        hdunum = ext+1
+    except TypeError:
+        extname=mks(ext).upper()
```
beckermr: Do we need to handle the case_sensitive keyword here?

fitsio/test.py
```diff
@@ -219,6 +219,7 @@ def setUp(self):
                      {'name':'empty','value':''},
                      {'name':'test2','value':'stuff','comment':'this is a string keyword'},
                      {'name':'dbl', 'value':23.299843,'comment':"this is a double keyword"},
+                     {'name':'edbl', 'value':1.384123233e+43,'comment':"double keyword with exponent"},
```
beckermr: Do we want to add a '1.38E+03' test case here?

fitsio/fitslib.py
```diff
@@ -4244,7 +4271,11 @@ def add_record(self, record_in):
 
             If the input is a card string, convert is implied True
         """
-        record = FITSRecord(record_in, convert=self.convert)
+        if isinstance(record_in,dict) and 'name' in record_in and 'value' in record_in:
+            record = {}
+            record.update(record_in)
+        else:
+            record = FITSRecord(record_in)
```
beckermr: A comment on what case this `if...else` is catching would be helpful.

fitsio/fitslib.py
```diff
@@ -4244,7 +4271,11 @@ def add_record(self, record_in):
 
             If the input is a card string, convert is implied True
         """
```
beckermr: It appears the doc string is out of date here.

fitsio/fitslib.py
esheldon: probably, thanks

fitsio/test.py
esheldon: for the zero in the 03?

fitsio/fitslib.py
esheldon: its just an optimization

fitsio/fitslib.py
esheldon: yep, thanks

fitsio/test.py
beckermr: And the capital E

fitsio/fitslib.py
beckermr: I’m not following but that’s ok.

fitsio/test.py
esheldon: well, cfitsio will write it how it writes it, putting a capital E in the python value won't change that.

BTW, it turns out the standard demands capital E in the header but cfitsio uses lower case, against the standard.

beckermr: ahhhh I thought these values went through the C code where you are testing for the capital E. I guess not. thanks!

PR #226, opened 2019-04-19
REF refactor into more files

beckermr:
This PR refactors the library into more files etc.

### DIFF
--- a/fitsio/hdu/base.py
+++ b/fitsio/hdu/base.py
@@ -1,0 +1,351 @@
+import copy
+import warnings
+
+from ..util import _stypes, _itypes, _ftypes, FITSRuntimeWarning
+from ..header import FITSHDR
+
+ANY_HDU = -1
+IMAGE_HDU = 0
+ASCII_TBL = 1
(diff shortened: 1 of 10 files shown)


fitsio/hdu/base.py
```diff
@@ -0,0 +1,351 @@
...
+        """
+
+        if isinstance(records_in, FITSHDR):
+            hdr = records_in
+        else:
+            hdr = FITSHDR(records_in)
+
+        if clean:
+            is_table = hasattr(self, '_table_type_str')
+            # is_table = isinstance(self, TableHDU)
```
beckermr: @esheldon This would cause a circular import in the new file structure. It also should probably be located in the TableHDU class and not here. Instead of doing that, I am simply looking for an attribute that is always set when a TableHDU class is made.

beckermr:
@esheldon This PR is ready for a good look. I refactored the code internally by splitting up a bunch of the classes and functions. The code also passes flake8 now except the tests, which I will do in another PR.

PR #244, opened 2019-05-03
return non-standard keyword val as a string

### LINKED ISSUE #243: trouble with NAN in headers
fitsio 1.0.3 crashes on FITS files with unquoted NAN in their header keywords, e.g.

```
DUSTA   =                  NAN / should be quoted but oops
```

Earlier versions of fitsio (e.g. 0.9.11) were more forgiving of this and would parse it as a string 'NAN'.

This header doesn't follow the FITS standard (the NAN should be in quotes), but it comes from IDL sxaddpar used by the SDSS/BOSS pipeline, and appears in SDSS data releases, e.g.

https://dr15.sdss.org/sas/dr15/eboss/spectro/redux/v5_10_0/6501/spPlate-6501-56563.fits

Would you consider updating fitsio to support this case again?

```
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-2-48c0e6c5fc48> in <module>
----> 1 hdr = fitsio.read_header('spPlate-6501-56563.fits', 0)

~/anaconda3/envs/fitsiotest/lib/python3.7/site-packages/fitsio-1.0.1-py3.7-macosx-10.7-x86_64.egg/fitsio/fitslib.py in read_header(filename, ext, extver, case_sensitive, **keys)
    188                     'hdu not found: %s (extver %s)' % (extname, extver))
    189 
--> 190     return FITSHDR(_fits.read_header(hdunum))
    191 
    192 

OSError: FITSIO status = 407: bad string to int conversion
Error in ffc2j evaluating string as a long integer: NAN
```

esheldon:
closes #243


sbailey:
Consider adding INF, NAN, NaN, -INF to test suite of real-world header keywords.

beckermr:
@esheldon and I were just talking about that. We don't have access to code that can write non-standard headers. Care to attach a small test file @sbailey ?

fitsio/test.py
```diff
@@ -525,6 +525,31 @@ def testHeaderFromCards(self):
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             """
```
beckermr: flake8 is going to die on this string...

esheldon: can I just add `# noqa` after ?

sbailey:
heading to a meeting, but will post a file later today

fitsio/test.py
beckermr: yup - or store the data as a file

beckermr: You can break it across lines too with `\`

esheldon:
@sbailey I added a test where I just put in a string representing a header.

PR #257, opened 2019-06-20
check LONGLONG first for int64 image dtype

### LINKED ISSUE #256: 64-bit integer images are not written correctly.
The following code shows the problem (with fitsio 1.0):
```
import numpy as np
import fitsio
arr = np.zeros(100, dtype=np.int64)
fitsio.write('test.fits', arr)
arr2 = fitsio.read('test.fits')
print(arr2.dtype)
```
And this returns `int32` when `int64` is expected.  Looking at the fits header shows that it was written with BITPIX = 32, so it's a problem on the writing.

esheldon:
closes #256


esheldon:
@erykoff check this

PR #259, opened 2019-07-16
ENH use explicit keywords

### LINKED ISSUE #258: move to explicit keyword arguments
We need to stop using `**kwargs` in so many spots. It is causing silent bugs for at least one user (me!).

beckermr:
This PR switches the code to using explicit keyword args. Closes #258

### DIFF
--- a/fitsio/hdu/base.py
+++ b/fitsio/hdu/base.py
@@ -22,28 +22,15 @@ class HDUBase(object):
     """
     A representation of a FITS HDU
 
-    construction parameters
-    -----------------------
+    parameters
+    ----------
     fits: FITS object
         An instance of a _fistio_wrap.FITS object.  This is the low-level
         python object, not the FITS object defined above.
     ext: integer
         The extension number.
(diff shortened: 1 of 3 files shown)


esheldon:
Perhaps we should keep the `**kw` in there and give a one-time warning if it contains anything.

This way codes that had bugs in a keyword (e.g. extname= on reading) will continue to work but will see a warning `unused keyword extname present; this will cause a crash in version XXXX of fitsio`

beckermr:
Yeah we can do that. However this is not a breaking change. Thus I am less motivated to be helpful.

beckermr:
I was thinking instead we bump the minor version to 1.1 to give people a chance to pin less if they need to

beckermr:
OK @esheldon. How strongly do you feel about keeping **kwargs with a warning (to be removed in a later version)?

I'd prefer an error to flush out silent bugs but I get that people might not want everything to break all at once.

beckermr:
Other than this, the PR itself is done.

PR #261, opened 2019-08-06
ENH update to cfitsio 3.470

### LINKED ISSUE #260: upgrade to cfitsio 3.470
It appears that cfitsio has released another version. We should upgrade.

at88mph:
Added cfitsio3470 and altered the setup.py script to use it.

Closes #260


beckermr:
This should replace the older version.

beckermr:
Also, we ship some patches to the library, some of which may have been pushed upstream. You will need to check those as well.

at88mph:
OK, changing this to a WIP for today.  Thanks.

cfitsio3470/docs/changes.txt
```diff
@@ -1,4 +1,93 @@
...
+  
+  - histo.c uses double precision internally for all floating point
+    binning; new double-precision subroutines fits_calc_binningd(),
+    fits_rebin_wcsd(), and fits_make_histd(); existing
+    single-precision histogram functions still work but convert values
+    to double-precision internally.
+
+  - new subroutine fits_copy_cols() / ffccls() to copy multiple columns
+  
+  - Fix in imcompress.c for HCOMPRESS and PLIO compression of unsigned
```
beckermr: There is a patch for this bug in particular that may need to be removed.

at88mph: OK.  Does the Travis build fail if the patches fail?

beckermr: Unclear. You will need to go through the patches by hand and check the source to make sure they still work. These patches are the least-worst solution.

at88mph:
Thanks for the feedback.  Do you want to manage the version in `setup.py` yourself?

beckermr:
I'm not sure what you mean. The cfitsio version or the fitsio version?

at88mph:
The `fitsio` version.  I would assume it would move to `1.0.6`.

beckermr:
ahhhhhh - we should hold off on that. There may be other PRs.

at88mph:
Thank you.  Does the `cfitsio` source contain the patched version or the pristine one?

beckermr:
The source folder in the repo should be pristine. It is then patched on the fly on install.

patches/putcols.c.patch
```diff
@@ -1,19 +0,0 @@
...
-+             if ( fits_use_standard_strings() ) {
-+                 if (snull[0] == ASCII_NULL_UNDEFINED) {
-+                     *buffer = ' ';
-+                 } else {
-+                     *buffer = '\0';
-+                 }
-+             } else {
-+                 *buffer = ' ';
-+             }
-+
```
beckermr: This patch is needed for string handling in python 3. We cannot remove it.

beckermr: unless the upstream file has been patched, which as far as I know it has not been.

at88mph: It doesn't apply to 3470.  Should it be adapted to it then?

at88mph: The line number was off by one.  Fixed now and re-added.

beckermr:
I double-checked the patches by hand and they all appear to apply properly to the new source code.

I think we are good to merge this. @esheldon do you have any thoughts?

esheldon:
I'm fine with it

PR #263, opened 2019-09-26
Fixes #262

### LINKED ISSUE #262: Slicing with scaling
I have a use case where I'm trying to perform some slicing:

```python
fits = fitsio.FITS('myfile.fits', 'r')
image_hdu = fits[1]
cutout = image_hdu[10:60]
```

The `myfile.fits` contains `SIP` entries with `BZERO`/`BSCALE` header cards.  I would like to slice on the pixels without having them scaled.  Can that be done?  

I think `cfitsio` supports it but I'm not sure if I can go that low level through `fitsio`.  Is there a solution with `numpy`'s `memmap`?

Thank you!

at88mph:
Added `ignore_scaling` flag to be set by the caller when reading in HDUs with `BZERO` and `BSCALE` headers set to values other than 0.0 and 1.0 respectively.

This PR is for reading HDUs only.  If the writing is needed, too, I will add it.

The `drvrnet.c` patch was added to fix an error during build.


at88mph:
Includes non-relevant items in PR.  Closing it.

PR #265, opened 2019-09-26
Fixes #264

### LINKED ISSUE #264: Dimension is not properly calculated when striding
```
import fitsio
hdu = fits[-1]
cut = hdu[2:20:2,:] # This is valid syntax, but the calculated dimensions in image.py are wrong as the step is not taken into account (i.e. assumes stop-start+1).
```

I think this is a bug since the `step` value is relevant in the calculation.

at88mph:


at88mph:
Includes non-relevant items in PR.  Closing it.

PR #269, opened 2019-09-27
Fixes #268

### LINKED ISSUE #268: Support flipping values along axis when slicing
```python
import fitsio
hdu = fits[-1]
cut = hdu[:,80:40] # This is valid syntax, but is explicitly denied in the image.py.
```

This could be enabled with a proper dimension calculated to provide this feature.

at88mph:


esheldon:
Unless I'm mistaken, I think you have changed the way slices work.  We support python slicing that is 0 offset and is exclusive and we translate it into cfitsio slices.

in testImageSlice you are comparing slices that are different in range by 1, implying you have changed it so that 1-offset slicing is used.

We definitely don't want to modify how the slicing works.

beckermr:
One other comment (and I am guilty of this too), I'd be great to have one PR per issue. This makes understanding diffs easier. No need to do that here, but def a good idea for the future.

beckermr:
Ahhhhh, your PRs are on top of the other ones in some cases. I thought I saw this code more than once....

at88mph:
Ah nuts.  I put them in different branches, but maybe my master one is getting mutated.  Sorry about that.

at88mph:
Actually, they are in different branches.  I think there is some overlap between my PRs given that I had everything in my master branch and had to split them out.  I may have messed it up a bit.

at88mph:
> Unless I'm mistaken, I think you have changed the way slices work. We support python slicing that is 0 offset and is exclusive and we translate it into cfitsio slices.
> 
> in testImageSlice you are comparing slices that are different in range by 1, implying you have changed it so that 1-offset slicing is used.
> 
> We definitely don't want to modify how the slicing works.

Meaning that if I request a slice from `3:10`, FITSIO will return an array of 6 instead of 7?

esheldon:
What I mean is that 3:10 returns an array of size 7

Also, this is zero-indexed rather than the cfitsio rule of being 1 indexed.  I think your new code would instead make it 1 offset right?

esheldon:
Can you alter the PR to preserve the zero offset slicing and python style `[start, end)` ?

at88mph:
I think so.  Would it alter the end result, though?  Are you referring to the `dimension` calculation?

Should I close this and try my PR again?  I put each of my PRs for the recent issues into separate branches to have a a single PR per issue, but they all seemed to converge into one for some reason.  I'm sorry about that.

esheldon:
Yes, it would alter the end result because we expect `hdu[3:6]` to return elements 3 4 and 5 but your new one would return elements 4 5 6, unless I am mistaken.

I must admit I do find the new PRs quite confusing, partly because they don't have descriptive names and also because they have common commits

at88mph:
Thanks.  I will close this and try it again.

esheldon:
Thanks Dustin.  I think these PR are going to be very useful so let's keep at it.

PR #271, opened 2019-10-09
Fixes the HISTORY keyword having a None value when read in.

at88mph:
Fixes #266 

This will check for a `HISTORY` keyword and treat it like a comment and call `add_string_to_dict` for the `comment` value.

A test has been added for it as well.

PR #273, opened 2019-10-09
Scaling ignore feature 262

at88mph:
This is to add a feature to the `base.py` to use the `ignore_scaling` flag on an HDU before reading the data from it.  Default value is `False` to preserve existing functionality.

This is useful when the `BZERO` AND `BSCALE` values are set to scale the pixels, but the user/caller wants to cut from the original pixels.

The API ice used like this:
```python
import fitsio

fits = fitsio.FITS('file.fits', 'r'). # File with BZERO=23.5 and BSCALE=0.4 for example.
hdu_1 = fits[1]
cut_1 = hdu_1[20:100,:]. # Will return scaled slice (BZERO+DAT*BSCALE)

hdu_1.ignore_scaling = True
cut_2 = hdu_1[20:100,:]. # Will return unscaled slice (BZERO=0.0, BSCALE=1.0)
```

**Note**:
In order for tests to pass, this relies on the `test.py` fixes from PR #271.

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -543,7 +543,7 @@ PyFITSObject_movabs_hdu(struct PyFITSObject* self, PyObject* args) {
 // get info for the specified HDU
 static PyObject *
 PyFITSObject_get_hdu_info(struct PyFITSObject* self, PyObject* args) {
-    int hdunum=0, hdutype=0, ext=0;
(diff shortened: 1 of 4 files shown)


esheldon:
Thanks Dustin.  Can you please also add a test that the values come back as expected?  I only see a test of the dtype

esheldon:
This looks good to me also and all tests pass.   This does point out, however, that we don't have any other tests with scaling.

PR #274, opened 2019-10-10
Add fix to allow striding.

### LINKED ISSUE #264: Dimension is not properly calculated when striding

at88mph:
Fixes #264 

Specifying a step in an image slice should shrink the dimension of the output.

```python
import fitsio
import numpy

fits = fitsio.FITS('file.fits', 'rw')
data = numpy.arange(10000).reshape(100,100)
fits.write(data)
hdu = fits[-1]
cut = hdu[:,10:50:8]. # Starting at pixel 10, step every 8th pixel until the 50th one along the second axis.
cut.shape
```

Will output:
```python
(100, 5)
```

Rather than:
```python
(100, 40)
```

This is useful when slicing a large portion of a larger image but want to skip some pixels to save space.

### DIFF
--- a/fitsio/hdu/image.py
+++ b/fitsio/hdu/image.py
@@ -25,6 +25,7 @@ from functools import reduce
 
 import numpy
 
+from math import floor
 from .base import HDUBase, IMAGE_HDU
 from ..util import IS_PY3, array_to_native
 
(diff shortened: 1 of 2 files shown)


esheldon:
I believe that should output 5
```python
>>> x = np.arange(100)
>>> x[10:50:8].size
5
```

at88mph:
Typo, sorry.  You're right.  Thanks.

esheldon:
This looks good.

PR #275, opened 2019-10-10
Support flipping along the axis when slicing

### LINKED ISSUE #268: Support flipping values along axis when slicing

at88mph:
Fixes #268 

This will likely conflict with PR #274 .

This feature enables a caller to slice along an axis in reverse.

```python
import fitsio
import numpy

fits = fitsio.FITS('outputfile.fits', 'rw', clobber=True)
dtype = numpy.int16
data = numpy.arange(100 * 200, dtype=dtype).reshape(100, 200)
fits.write_image(data)
hdu = fits[-1]
rdata = hdu[:,130:70]
rdata.shape # Will be (100, 60)

ndatacut = data[:,130:70:-1]. # Equivalent to numpy directly.
ndatacut.shape # Will be (100, 60)
```

### DIFF
--- a/fitsio/test.py
+++ b/fitsio/test.py
@@ -1071,17 +1071,18 @@ DATASUM =                      / checksum of the data records\n"""
                         "Data are not the same (Expected shape: {}, actual shape: {}.".format(
                             expected_data.shape, rdata.shape))
 
-                try:
-                    hdu[:,90:60:4]  # Positive step integer does not follow numpy syntax.
-                    raise AssertionError('Should raise ValueError.')
-                except ValueError as value_error:
-                    self.assertEqual('{}'.format(str(value_error)), 'slice steps must be < 0 when stop < start')
+                
+                rdata = hdu[:,90:60:4]  # Positive step integer with start > stop will return an empty array
+                expected_data = numpy.empty(0, dtype=dtype)
+                numpy.testing.assert_array_equal(expected_data, rdata,
+                        "Data are not the same (Expected shape: {}, actual shape: {}.".format(
+                            expected_data.shape, rdata.shape))
 
-                try:
-                    hdu[:,60:90:-4]
-                    raise AssertionError('Should raise ValueError.')
-                except ValueError as value_error:
-                    self.assertEqual('{}'.format(str(value_error)), 'slice steps must be >= 1 when start < stop')
+                rdata = hdu[:,60:90:-4]  # Negative step integer with start < stop will return an empty array.
+                expected_data = numpy.empty(0, dtype=dtype)
+                numpy.testing.assert_array_equal(expected_data, rdata,
+                        "Data are not the same (Expected shape: {}, actual shape: {}.".format(
+                            expected_data.shape, rdata.shape))
         finally:
             if os.path.exists(fname):
                 os.remove(fname)
--- a/fitsio/test.py
+++ b/fitsio/test.py
@@ -1062,7 +1062,7 @@ DATASUM =                      / checksum of the data records\n"""
                         "Data are not the same (Expected shape: {}, actual shape: {}.".format(
                             expected_data.shape, rdata.shape))
 
-                rdata = hdu[:,130:70:6]
+                rdata = hdu[:,130:70:-6]
 
                 # Expanded by two to emulate adding one to the start value, and adding one to the calculated dimension.
                 expected_data = data[:,130:70:-6]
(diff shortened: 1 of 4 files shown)


at88mph:
I fixed a couple of conflicts.  Thank you!

esheldon:
I'm finally looking at this in detail now.

Why is the striding in fits having the opposite sign from the numpy version?

at88mph:
Thanks.

When the `x` value is larger than the `y` value of the range, then we need to explicitly tell `numpy` to start counting backwards:

```python
import numpy
data = numpy.arange(100).reshape(10,10)
data[:, 6:5] # Returns an empty array.
>>> array([], shape=(10, 0), dtype=int64)
data[:,6:5:-1]
>>> array([[ 6],
       [16],
       [26],
       [36],
       [46],
       [56],
       [66],
       [76],
       [86],
       [96]])
```

esheldon:
What I'm asking is why you're new code doesn't use the numpy syntax?  I
intended the slicing to mirror numpy and i would like to continue that

at88mph:
Sorry, I'm not following.  The new code just builds on what's already there.  Are you suggesting that it be left up to the caller to appropriately set the `step` value?  So the caller would write:

```python
import fitsio
import numpy

hdu = fits[0]
cut = hdu[:,60:45:-1]
```

Instead of just writing:

```python
...
cut = hdu[:60:45]
...
```

at88mph:
Leaving it to the caller may be better anyway.  In the current PR, the caller would have to call with a different step value like so:

```python
import fits
import bumpy

hdu = fits[0]
cut = hdu[:,60:45:4]. # Use a step value of 4, which becomes -4.
```

That may not be intuitive.  If we did do that, though, would `fitsio` have to assert that if the `stop` value is less than the `start`, then the `step` value is negative?

If I'm misreading, please let me know.

at88mph:
Update to enforce `numpy` syntax when slicing.  Callers _must_ specify a negative step value (or omit it) when slicing and flipping along an axis.

at88mph:
Can someone restart the TravisCI build?  It looks like one of the builds failed to download dependencies due to a network timeout.

beckermr:
Feel free to push random commits to restart things.

fitsio/test.py
```diff
@@ -1062,14 +1062,26 @@ def testReadFlipAxisSlice(self):
...
 
                 # Expanded by two to emulate adding one to the start value, and adding one to the calculated dimension.
                 expected_data = data[:,130:70:-6]
 
                 numpy.testing.assert_array_equal(expected_data, rdata,
                         "Data are not the same (Expected shape: {}, actual shape: {}.".format(
                             expected_data.shape, rdata.shape))
+
+                try:
+                    hdu[:,90:60:4]  # Positive step integer does not follow numpy syntax.
```
esheldon: What does this comment mean, that it does not follow numpy syntax?

esheldon: with a numpy array, the above would return an empty array

fitsio/test.py
```diff
@@ -1062,14 +1062,26 @@ def testReadFlipAxisSlice(self):
...
+                    raise AssertionError('Should raise ValueError.')
+                except ValueError as value_error:
+                    self.assertEqual('{}'.format(str(value_error)), 'slice steps must be < 0 when stop < start')
+
+                try:
+                    hdu[:,60:90:-4]
```
esheldon: with a numpy array, the above would return an empty array

fitsio/test.py
at88mph: That's correct, it does return an empty array.  `numpy` syntax wants a negative number when `x` < `y` for a slice.  Given that it will always return an empty array in this situation, I thought it better to inform the user.

fitsio/test.py
at88mph: That's correct, it does return an empty array.  `numpy` wants a negative number when `x` < `y` for a slice.  Given that it will always return an empty array in this situation, I thought it better to inform the user.

fitsio/test.py
esheldon: But are you not raising an error for that case?

at88mph: Right, by inform I mean it won't return what you think it will return, so that may be reason enough to die.  

In `CFITSIO`, the syntax is to use a positive integer for the `step` value for this use case, but with the `numpy` syntax, it requires a negative one, so there may be confusion from the caller's standpoint.

esheldon: I think maybe the right thing to do would be to return the empty array rather than raising an exception.  We have the type info available, so we could do that.

If you want we could do it in a different PR.  Does anyone else have an opinion on it?

at88mph: Sure.  Lines 280 to 283 in `image.py` would perhaps instead look like:
```python
            if (step > 0 and stop < start) or (step < 0 and start < stop):
                return numpy.empty(0, dtype=npy_dtype)
```

Then the tests could just check for an empty array.

esheldon:
Sorry for the long delays @at88mph 

I was very busy and then took a long vacation. I think this PR looks good.

The changes do not pass flake8 tests but we can clean that up in a separate PR

I'll go ahead and merge

PR #278, opened 2019-11-16
Fix error reading with an empty rows argument.

rainwoodman:
Also see: https://github.com/bccp/nbodykit/issues/606


esheldon:
This looks good to me.

PR #281, opened 2019-12-04
update changes and flake8 tests on image.py

esheldon:


esheldon:
The changes.txt looks a bit off; some of the changes put in 1.05 I think were after that release right?

Issue #150, opened 2018-02-15
writing one compressed image forces subsequent ones to be compressed


sbailey:
Writing a compressed image to an open FITS file causes all subsequent images to be compressed.  If the images are written individually using `fitsio.write` then the expected behavior works where individual images are compressed only if requested.  Example:
```python
import numpy as np
import fitsio

x = np.random.uniform(0,1,size=(5,8))

with fitsio.FITS('blat.fits', mode='rw', clobber=True) as fx:
    fx.write(x, extname='A')
    fx.write(x, extname='B', compress='gzip')
    fx.write(x, extname='C', compress=None)  #- OOPS: will be compressed too   

fitsio.write('foo.fits', x, extname='A', clobber=True)
fitsio.write('foo.fits', x, extname='B', compress='gzip')
fitsio.write('foo.fits', x, extname='C', compress=None) #- Correctly will not be compressed

for filename in ['blat.fits', 'foo.fits']:
    print('Reading {}'.format(filename))
    fx = fitsio.FITS(filename)
    for hdu in fx:
        print('  {:8s} {}'.format(hdu.get_extname(), hdu.is_compressed()))
```

Produces:
```
Reading blat.fits
  A        False
  B        True
  C        True
Reading foo.fits
  A        False
  B        True
  C        False
```
I was expecting HDU C of blat.fits to be uncompressed, just like HDU C of foo.fits.

esheldon:
thanks for the report.  Do you have a fix for this in mind?

sbailey:
From a quick read of the code I don't see why it isn't working and was hoping that you would know faster.  i.e. I don't see the code ignoring the `compress` keyword so I don't have a fix in mind without needing to do more detailed debugging.

esheldon:
@sbailey I have a fix for this in PR #218

PR #285, opened 2020-01-16
Fpack in filename

### LINKED ISSUE #283: Fpack compression stopped working?
Hi,

In fitsio 0.9.12, this code:
```
import fitsio
import numpy as np
img = np.zeros((1000,1000), np.float32)
fitsio.write('/tmp/test.fits.fz[compress R 100,100; qz -1e-4]', img, clobber=True)
```
produces a FITS file that is fpack compressed.

In fitsio 1.0.5, the same code results in a non-compressed file.

Also, `clobber=True` doesn't work in the latter case.

This is in python3, with fitsio installed on a fresh ubuntu 18.04 container using
pip3 install fitsio==0.9.12
or
pip3 install fitsio==1.0.5

cheers,
--dustin

esheldon:
closes #283

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -1347,7 +1347,8 @@ PyFITSObject_create_image_hdu(struct PyFITSObject* self, PyObject* args, PyObjec
         }
 
         // 0 means NOCOMPRESS but that wasn't defined in the bundled version of cfitsio
-        if (comptype >= 0) {
+        // if (comptype >= 0) {
+        if (comptype > 0) {
             // exception strings are set internally
             if (set_compression(self->fits, comptype, tile_dims_obj, &status)) {
                 goto create_image_hdu_cleanup;
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -466,8 +466,8 @@ class FITS(object):
         if self.mode in [READWRITE, 'rw']:
             if clobber:
                 create = 1
-                if os.path.exists(filename):
-                    os.remove(filename)
+                if filename[0] != '!':
+                    filename = '!' + filename
             else:
                 if os.path.exists(filename):
                     create = 0


beckermr:
This change will reintroduce the bug that it solved here: https://github.com/esheldon/fitsio/issues/150

beckermr:
Or at least we should test.

esheldon:
that fix is what introduced this bug :)

esheldon:
looking at this, it is the convention for cfitsio that compression stays on until you turn it off, so it was doing the expected thing before, from a historical perspective

esheldon:
@sbailey

beckermr:
> that fix is what introduced this bug :)

So we have to turn it off in the code in order to not reintroduce the issue where if you write one image, the subsequent ones get compressed too. Otherwise we need to reopen #150.

esheldon:
right, these are totally incompatible basically.  If I had looked into #150 more I would have seen that we should not have "fixed" it

esheldon:
but... it does mean we should add the ability to turn off compression when writing a new hdu

beckermr:
I think we have to do it automatically somehow...

moustakas:
Thanks @esheldon! I've tested and this PR works. I suspect @sbailey won't care if we defer a fix for #150 to a different PR, so this can get merged. (And if you could cut a `1.0.6` tag...)

esheldon:
@beckermr  I think I'm not being clear.  In cfitsio land you have to turn it off "by hand", and by putting in the fix for #150 we broke that, and broke the "compression in filename" as well.

Basically that "bug" was a "feature" and we screwed up by fixing it.

beckermr:
Ahhhh. If we want to emulate the cfitsio interface, then great. We should provide a hook to turn it off at the python level.

esheldon:
That's the idea, but I would like to get feedback from @sbailey

sbailey:
If fitsio provides a way to turn compression back off after writing one compressed image, I think that would be sufficient.  It was the enforced "once compressed always compressed" feature/bug that was problematic.  Our particular use case was that we wanted to write integer images (masks) as compressed since it is lossless, but did *not* want to write subsequent float images as compressed because that was lossy, and then tripped over this problem.

beckermr:
Sounds good. A context manager for this might ideal

```python

import fitsio

f = fitsio.FITS(...)

with f.compressed():
    # do something with f

```

dstndstn:
echoing @moustakas , we would super appreciate it if you could merge and tag a release with this fix!

esheldon:
should we wait until the option to turn the compression off is implemented?

moustakas:
Restoring the compression option functionality is far higher priority for
DESI right now.

PR #286, opened 2020-01-25
Fix bug when reading slice with step, but no start/stop

rmjarvis:
I ran into this bug when adding a feature to TreeCorr that would read only every nth row in the input catalog.  It worked when start or stop were specified, but not when reading from the whole range.


esheldon:
LGTM

esheldon:
thanks mike

PR #299, opened 2020-04-28
write long long rather than long

esheldon:


esheldon:
travis is not triggering

at88mph:
Awesome, thanks.  I realize this is likely a bad database ID that was used to generate the file or something, so much obliged.

esheldon:
huh... travis is actually running, it just isn't showing up here

at88mph:
I've noticed that GitHub now prefers you to create their Actions rather than use Travis these days.

esheldon:
there is now an incident for the CI problem:  https://www.githubstatus.com/

esheldon:
they claim to have fixed it but it is not fixed

I'm starting to think there is a problem with travis because, for example, mac os builds are not starting

esheldon:
It ran and succeeded, gonna merge

PR #300, opened 2020-04-28
BUG disallow getitem when file created but no data

### LINKED ISSUE #290: empty image hdus
This code block did some unexpected things

```
with fitsio.FITS("test.fits", mode='rw', clobber=True) as fp:
    if fp[0].has_data():
        fp[0].write(
            np.zeros((10000, 10000)))
    else:
        fp.write(np.zeros((10000, 10000)))
```

this works

```
with fitsio.FITS("test.fits", mode='rw', clobber=True) as fp:
    fp.write(np.zeros((10000, 10000)))
```

esheldon:
When we created the file and haven't created anything, we will not

closes #290

### DIFF
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -474,6 +474,7 @@ class FITS(object):
                 else:
                     create = 1
 
+        self._did_create = (create == 1)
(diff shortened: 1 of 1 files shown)


esheldon:
still no travis

esheldon:
travis did run, just didn't show up here

PR #301, opened 2020-05-21
Remove extraneous debugging print statements

erykoff:


esheldon:
argh, travis is not running

PR #311, opened 2020-11-12
don't remove BLANK cards in header clean

### LINKED ISSUE #303: Allow BLANK header cards
The FITSHDR clean() method removes `BLANK` cards. I don't believe that it should. Unlike NAXISx and BITPIX, the value cannot be recovered from integer image data, which is the only data which BLANK is valid for. 

All the special treatment of BLANK in cfitsio has to do with compression, but that only comes into play when doing lossy compression of integer data. In all other cases BLANK is left alone. [I think the standard says that BLANK _should_ also be used in the case of lossy integer compression, but that ZBLANK is OK.]

I suppose clean() could remove BLANK from headers for floating-point images, but that feels like the wrong place to do that.

esheldon:
closes #303

### DIFF
--- a/fitsio/header.py
+++ b/fitsio/header.py
@@ -222,7 +222,7 @@ class FITSHDR(object):
             'SIMPLE', 'EXTEND', 'XTENSION', 'BITPIX', 'PCOUNT', 'GCOUNT',
             'THEAP',
             'EXTNAME',
-            'BLANK',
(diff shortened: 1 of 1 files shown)


esheldon:
@CraigLoomis does this work for you?

CraigLoomis:
Yup -- that is all that I think needs to be done.

PR #312, opened 2020-11-12
Store card string

### LINKED ISSUE #293: Header card_string
Hello!

I have been updating to the last version of fitsio lately and I came across a change with respect to the previous version I have been using for a while (0.9.11):
When using the function class FITS and its method fits_header you used to provide a list of dictionary for each entry in the fits header with a 'card_string' keyword [along with name, comment and value]. I was using this because I could catch a word used ['HIERARCH'] in the fits header keywords that for some reason is not seen.

Do you know if there is a way to mock the behavior of 'card_string' in the new version?

esheldon:
closes #293 

Store `card_string` in the header entries read from disk.

These can contain incomplete data, so for repr we use a fake card that can be longer

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -4329,6 +4329,8 @@ PyFITSObject_read_header(struct PyFITSObject* self, PyObject* args) {
         }
 
         dict = PyDict_New();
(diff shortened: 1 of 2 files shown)


esheldon:
for the last push, I had accidentally left in a debugging version of the entry

PR #314, opened 2020-12-08
Adding Power support(ppc64le) with ci and testing to the project for architecture independent

asellappen:
I am working for IBM to port cpu arch ppc64le for open sources.

This PR adds CI support for the IBM Power Little Endian (ppc64le) architecture. The idea is to ensure that the builds on this architecture are continuously tested along with the Intel builds (amd64) as this is part of the ubuntu distro 
on that architecture as well and detecting (and fixing) any issues or failures early would help to ensure that we are always up to date.

This helps us simplify testing later when distributions are re-building and re-releasing,We typically build applications for customers and 
ISVs, and while we don't use this package directly,we do count on all of the packages in debian/ubuntu to build other packages. So we more likely have this as a second or 
third level dependency and couldn't tell you explicitly which features we use or our usage model,For more info tag @gerrith3.

Please help to verify and merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ os:
     - osx
 
 language: c
+arch:
+    - amd64
+    - ppc64le
 
 # use the containers for a faster build
 sudo: false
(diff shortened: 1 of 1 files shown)


beckermr:
This job will consume more of the limited amount of free CI time provided by travis, which is the only provider I know for ppc64le builds. Can IBM provide the CI time for this changed?

asellappen:
Thanks for taking up .
We can get free credits from travis by requesting for more build minutes , Please create a request at https://support.travis-ci.com/hc/en-us .
We are working on many packages and getting used  all CI times very frequently  during these packaging build , Whenever we require more credits then we request using the above url  to raise a ticket, and getting more CI time easily.

beckermr:
Unfortunately, this places a burden on package maintainers to negotiate with TravisCI on a regular basis for credits. Can you provide a solution that eliminates this extra work? I plan on moving this repo to GitHub Actions anyway in order to avoid having to do this.

esheldon:
Can we use travis for the ppc64le builds and  github actions for the  rest?  That would reduce total usage of travis

beckermr:
We can but once we run out of minutes, unless we take the time to bother travis, the ppc64le tests will not run. We get a fixed number of free minutes and then we will be out. I don't think we want to take on the maintenance burden of having to bother travis for minutes on a regular basis to run tests on this alternate platform.

gerrith3:
@beckermr How many credits/minutes do you think you'd need for the next, say, year?  Submit that once, let me know and I'll escalate it.  @esheldon and yes, you can split CI's, and use GHA + Travis CI, each for what they are best at.

beckermr:
I've enabled ppc64le on master. @esheldon will need to email travis since the CI is under his account. I'd request 100,000 credits.

PR #319, opened 2021-02-05
Replace depreacted numpy.bool with numpy.bool_

erykoff:


beckermr:
Will this work with older numpy versions too?

beckermr:
Any comments @esheldon?

esheldon:
I think the recommendation is to use regular `bool` for this, e.g.
```python
np.zeros(num, dtype=bool)
```

esheldon:
https://numpy.org/devdocs/release/1.20.0-notes.html

See "deprecated".  It recommends using the second column rather than the third

erykoff:
I'm not sure if `numpy.bool_` vs `bool` matters much, but either will work.  I also forgot to update `table.py`.

erykoff:
I also confirmed that you can do dtype equivalence with `bool`, `numpy.bool_` or `numpy.bool` (the final one with a deprecation warning on 1.20).

esheldon:
any interest to add a bool test?  If not that's ok

erykoff:
Wasn't planning on it.  It's that right now whenever I see a warning here I do a drive-by PR to fix things (including, of course, my own code...)

beckermr:
I'm all for drive-by PRs. They make the world a better place.

Issue #298, opened 2020-04-12
Impossible to pip install


londumas:
Running pip install on the latest version crashes.
`python -m pip install fitsio --upgrade`, but not on version 1.0.5.
```
Collecting fitsio
  Using cached https://files.pythonhosted.org/packages/9c/7d/99906853351108cd5abea387240b5b58109a91e349f0ae22e33c63969393/fitsio-1.1.1.tar.gz
Requirement already satisfied, skipping upgrade: numpy in <HOME>.local/lib/python3.7/site-packages (from fitsio) (1.18.2)
Building wheels for collected packages: fitsio
  Running setup.py bdist_wheel for fitsio ... error
  Complete output from command /usr/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-install-f3s2mfyj/fitsio/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /tmp/pip-wheel-f4wmuk84 --python-tag cp37:
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.linux-x86_64-3.7
  creating build/lib.linux-x86_64-3.7/fitsio
  copying fitsio/header.py -> build/lib.linux-x86_64-3.7/fitsio
  copying fitsio/fitslib.py -> build/lib.linux-x86_64-3.7/fitsio
  copying fitsio/__init__.py -> build/lib.linux-x86_64-3.7/fitsio
  copying fitsio/test.py -> build/lib.linux-x86_64-3.7/fitsio
  copying fitsio/util.py -> build/lib.linux-x86_64-3.7/fitsio
  creating build/lib.linux-x86_64-3.7/fitsio/hdu
  copying fitsio/hdu/base.py -> build/lib.linux-x86_64-3.7/fitsio/hdu
  copying fitsio/hdu/image.py -> build/lib.linux-x86_64-3.7/fitsio/hdu
  copying fitsio/hdu/__init__.py -> build/lib.linux-x86_64-3.7/fitsio/hdu
  copying fitsio/hdu/table.py -> build/lib.linux-x86_64-3.7/fitsio/hdu
  running build_ext
  checking file build/temp.linux-x86_64-3.7/cfitsio3470/putcols.c
  Hunk #1 succeeded at 157 (offset -1 lines).
[...]

PR #322, opened 2021-04-13
Add support for automatically building wheels on Linux and Mac

dfm:
This PR adds a GitHub actions workflow that uses [cibuildwheel](https://github.com/joerick/cibuildwheel) to automatically build binary wheels for the standard Linux and Mac platforms. These wheels could be uploaded to PyPI manually or uploaded automatically (this is what I do with most of my projects). This would help with issues like #298 (because it packages all of the dependencies) and generally speed up pip-based dependency workflows, but some users might prefer to tune their cfitsio, and it adds some developer overhead, so you should definitely feel free to veto/ignore/etc.!

Also, I'll be happy to squash/rebase to clean up the history if this is something that you're interested in merging.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -17,26 +17,16 @@ jobs:
       matrix:
         os: 
           - ubuntu-20.04
-#           - windows-2019
           - macos-10.15
     steps:
       - uses: actions/checkout@v2
-        with:
-          submodules: true
-          fetch-depth: 0
-      
-      - uses: ilammy/msvc-dev-cmd@v1
-        if: runner.os == 'Windows'
-        
       - uses: joerick/cibuildwheel@v1.10.0
         env:
           CIBW_BUILD: "cp3?-*"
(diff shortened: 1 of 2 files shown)


beckermr:
I don't personally want to maintain this, but I'll leave the final say to @esheldon.

esheldon:
Thanks @dfm 

I don't know anything about wheels.  I certainly could not maintain this.

How does this affect publishing to pypi or conda forge?

dfm:
@esheldon, @beckermr: Thanks - I would find this a useful feature since I mostly use pip to manage my dependencies rather than conda, but I don't want to add extra maintenance burden! I think probably the best bet is to just close this if you aren't feeling comfortable with such a feature. Thanks for the consideration!

@esheldon: to answer your questions: I use a workflow like this to automatically publish to PyPI (both source and binary builds) on a tagged GitHub release ([for example](https://github.com/exoplanet-dev/exoplanet-core/blob/5b9a3497ae644cfbc69cccf8ce1b71d2dc76c730/.github/workflows/wheels.yml)). It would not have any affect on conda/conda-forge whatsoever.

Issue #325, opened 2021-08-25
Regression from 1.1.2 to 1.1.3: "keyword not found in header"


dstndstn:
Hi,

I'm trying to read a Pan-STARRS DR2 stacked image and while 1.1.2 works, 1.1.3 and 1.1.4 fail with:

```
  File "/Users/dstn/legacypipe/py/legacypipe/image.py", line 1105, in read_image_header
    self._image_header = self.read_image_fits()[self.hdu].read_header()
  File "/usr/local/lib/python3.8/site-packages/fitsio/hdu/base.py", line 343, in read_header
    return FITSHDR(self.read_header_list())
  File "/usr/local/lib/python3.8/site-packages/fitsio/hdu/base.py", line 358, in read_header_list
    return self._FITS.read_header(self._ext+1)
OSError: FITSIO status = 202: keyword not found in header
```

A small subimage reproduces the error:
http://broiler.astrometry.net/~dstn/temp/sub.fits
ie,

```
> python3 -c "import fitsio; fitsio.read_header('/tmp/sub.fits')"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.8/site-packages/fitsio/fitslib.py", line 238, in read_header
    return FITSHDR(_fits.read_header(hdunum))
OSError: FITSIO status = 202: keyword not found in header
```

These images don't have an EXTNAME card, but that doesn't seem to be the issue.

fitsverify doesn't complain about the image.

It does have HIERARCH cards.

esheldon:
The pdf https://fits.gsfc.nasa.gov/registry/hierarch/hierarch.pdf makes it clear that there should be no "." in these names.

So this is a malformed header. Did you identify what it was between 1.1.2 and 1.1.3 that causes it to fail?

esheldon:
Ahh, it is because we introduced the replacement of junk characters in header names :)

esheldon:
That was a serious issue because not doing so caused seg faults.

Maybe its a book keeping issue: the character gets replaced but then this is not understood somewhere down stream.

dstndstn:
Yes, that is exactly what happens, the keyword gets altered during header parsing, and then later during header parsing it looks up the value for that keyword using the original name, which then isn't found.

As for whether these headers are legal: in the HIERARCH documentation I guess you're referring to this paragraph,
```
[...]

PR #326, opened 2021-08-26
don't run convert_keyword_to_allowed_ascii for HIERARCH header cards

dstndstn:
To address #325


esheldon:
The pdf https://fits.gsfc.nasa.gov/registry/hierarch/hierarch.pdf makes it clear that there should be no "." in these names.

So this is a malformed header.  Did you identify what it was between 1.1.2 and 1.1.3 that causes it to fail?

beckermr:
@dstndstn This one can be closed when you are ready. Happy to do it myself too.

dstndstn:
Yes, thank you!

dstndstn:
solved by #332 instead.

PR #329, opened 2021-09-07
for bad key converted to ascii, return string values

esheldon:
closes #325 

this avoids a potential seg fault in fits_read_key

maybe should send a bug report on fits_read_key for this

PR #302, opened 2020-06-02
convert header garbage to ascii

esheldon:
For non-keywords the non-ascii characters are replaced
with '?'

For keywords some attempt is made to produce a new keyword that would be
valid, although this is not checked.

closes #295

PR #330, opened 2021-09-07
BUG fix latest segfault

beckermr:
This PR fixes the latest segfault for headers.


dstndstn:
The bug from #302 is fixed, but #325 still fails
```
> python -c "import fitsio; print(fitsio.__version__); fitsio.read_header('/tmp/bad-teeny.fits'); print('1'); fitsio.read_header('/tmp/tiny.fits')"
1.1.4
1
  File "<string>", line 1, in <module>
  File "/Users/dstn/fitsio/fitsio/fitslib.py", line 238, in read_header
    return FITSHDR(_fits.read_header(hdunum))
OSError: FITSIO status = 202: keyword not found in header

```

beckermr:
Yup that is to be expected. We will need this and the PR from @esheldon.

beckermr:
I've pushed the changes from @esheldon into this branch. Can you try again?

beckermr:
Both of the files work for me now. We need to write a test for the segfault.

dstndstn:
I confirm that both work for me!

dstndstn:
(though the header changes from "FRA.ZP" to "FRAKZP"...)

beckermr:
Right. We'll need to fix that in another PR.

beckermr:
I've checked that the test I added fails without this PR.

beckermr:
@esheldon I merged your PR #329 into this PR and added a segfault bug. I think both should be merged and then we can work on parsing to something better than we are now.

beckermr:
@dstndstn Are you ok if we ship the contents of the header you provided as part of the fitsio test suite under its license?

dstndstn:
Yes, that's fine, thank you for asking (and for all this work!).

esheldon:
If I followed the discussion, we will discuss how to replace JUNK___ in another issue/PR.

If so then this is ready to go

esheldon (approved):

PR #332, opened 2021-09-08
ENH clean up header handling

### LINKED ISSUE #331: change filling of bad characters in header keywords
currently filling with `JUNK____....`

beckermr:
This PR cleans up header handling. It does a few things:

1. When reading hierarch header entries, it allows any ascii chars.
2. It replaces junk it reads with `_` instead of using `JUNK__`.
3. It has tests for reads of total junk.
4. It raises an error for writing header keys with `?`, `*`, or `#` which appear to generate fits files that cannot be read.

superceeds #326
closes #331


beckermr:
@esheldon @dstndstn I think this has the rest of the work related to header parsing. Appears to pass all tests. If you all can try it on various files you have around to test it, that'd be great!

beckermr:
Also pardon the whitespace changes in C. My editor removes them automatically.

fitsio/hdu/base.py
```diff
@@ -4,6 +4,7 @@
 from ..util import _stypes, _itypes, _ftypes, FITSRuntimeWarning
 from ..header import FITSHDR
 
+VALID_HDR_CHARS = ("-", "_", ".")
```
beckermr, line 7:
```suggestion
```

beckermr:
Note that internall cfitsio appears to choke on using `?` for the weird ascii as opposed to other things like `@` or `<`.

beckermr:
Since these three templare characters write files that cannot be read, I am raising an error if the user attempts to write them.

fitsio/test.py
```diff
@@ -660,25 +660,23 @@ def testHeaderJunk(self):
...
         fname=tempfile.mktemp(prefix='fitsio-HeaderJunk-',suffix='.fits')
         try:
-            with open(fname,'w') as fobj:
+            with open(fname,'wb') as fobj:
                 fobj.write(data)
 
             h = fitsio.read_header(fname)
-            self.assertEqual(h['bar'],'NAN', "NAN garbage")
-            self.assertEqual(h['biz'],'NaN', "NaN garbage")
+            self.assertEqual(h['bar_'],'NAN', "NAN garbage")
```
esheldon, line 671: can you add comments here explaining the `_` is expected?

fitsio/hdu/base.py
```diff
@@ -320,6 +321,12 @@ def write_keys(self, records_in, clean=True):
             if name is not None:
                 name = name.upper()
 
+                if any(c in INVALID_HDR_CHARS for c in name):
```
esheldon, line 324: this is potentially slow in python when writing big headers.  We should test to see if this needs optimization, e.g. compiled regexp or check in C

fitsio/test.py
```diff
@@ -660,25 +660,23 @@ def testHeaderJunk(self):
...
             self.assertEqual(h['boo'],'-INF', "-INF garbage")
             self.assertEqual(h['quat'], '', 'blank')
             self.assertEqual(h['quip'], '1.0', '1.0 in quotes')
             self.assertEqual(h['quiz'], ' 1.0', '1.0 in quotes')
-            self.assertEqual(h['quil'], 'NaN', 'NaN in quotes')
-            # note replacement of junk '.' in index 2 N from
-            # index 2 of JUNK___
-            self.assertEqual(h['qund'], 'Inf', 'Inf in quotes')
+            self.assertEqual(h['qui__'], 'NaN', 'NaN in quotes')
+            self.assertEqual(h['qu.@d'], 'Inf', 'Inf in quotes')
```
esheldon, line 679: explain why this one went through unchanged

esheldon (changes requested):
This looks good, have some minor comments and requests

beckermr:
@esheldon This is ready for another look!

esheldon (approved):

esheldon:
@dstndstn do you want to try this out on some of your real world data before we merge?

dstndstn:
Works for my cases!

beckermr:
I have one more idea here. Let's hold off on merging for just a bit.

fitsio/fitsio_pywrap.c
```diff
@@ -104,22 +104,23 @@ static int convert_keyword_to_allowed_ascii(char* str) {
...
-                str[i] = 'U';
-            } else if (i==2) {
-                str[i] = 'N';
-            } else if (i==3) {
-                str[i] = 'K';
-            } else {
-                str[i] = '_';
-            }
+            str[i] = '_';
+            // if (i==0) {
```
esheldon, line 113: remove this commented code

esheldon (changes requested):

beckermr:
@dstndstn @esheldon can you all try this out on weird data you have one more time?

The parsing now reads as much as it can without conversion.

fitsio/fitsio_pywrap.c
```diff
@@ -4351,7 +4391,13 @@ PyFITSObject_read_header(struct PyFITSObject* self, PyObject* args) {
             add_string_to_dict(dict, "comment", scomment);
 
         } else {
-            was_converted = convert_keyword_to_allowed_ascii(keyname);
+
+            if (is_hierarch) {
```
esheldon, line 4395: why are these treated differently?  Can you please add a comment?

esheldon (changes requested):
Just to be sure, are we testing a case that would have seg faulted before?

fitsio/fitsio_pywrap.c
beckermr: Yep. I will add a comment. Here is the logic

If a key is HIERARCH, then any ASCII character except `*?#` is allowed. If we don't do any conversion, then we can use the FITS key read functions, otherwise we have to write the value as a string.

If a key is not HIERARCH, then we have more cases. If a key has non-allowed chars for standard fits header keys, then we cannot use the FITS key read functions. So we track that. Also, if a key has non-ascii characters, then we need to convert it before we write it to the python dict. 

All of the if blocks here do that logic. I will add comments.

dstndstn:
This works well for the original Pan-STARRS image I reported problems with (yay!).

beckermr:
ok @esheldon comments added - back to you!

beckermr:
> Just to be sure, are we testing a case that would have seg faulted before?

yes - we are testing stuff that would have segfaulted. I am not sure if it is the same case, but we have weird non-ascii stuff in our junk headers now.

esheldon (approved):

PR #338, opened 2021-11-24
BUG fix list assignment index out of range (#337)

### LINKED ISSUE #337: Updating an existing record in an image header raises an exception
I wish to update a existing record but an error occurs : `IndexError: list assignment index out of range` on self._record_list

In my opinion, this problem is due to a previous record delete. I'm trying to remove some keywords before updating an existing record. The update raises the exception.

Here, my code to reproduce my problem with a fits file as input : [fitsio_headers.txt](https://github.com/esheldon/fitsio/files/7570211/fitsio_headers.txt)
I'm just reading the first image extension header, adding a new record, deleting some keywords and overriding the previous record.
I've got the exception if I use the delete function and retrieve the same FITSHDR reference. However, if I build a new FITSHDR instance and keep only what I want, that works.

Perhaps I'm simply doing something wrong by delete the records.

I'm using  fitsio 1.0.5.

ussegliog:
This PR fixes #337 and adds a new unit test.

Closes #337

### DIFF
--- a/fitsio/header.py
+++ b/fitsio/header.py
@@ -202,8 +202,17 @@ class FITSHDR(object):
         else:
             if name in self._record_map:
                 del self._record_map[name]
+                # Store current index value
+                cur_index = self._index_map[name]
+                # Delete in index map
+                del self._index_map[name]
                 self._record_list = [
                     r for r in self._record_list if r['name'] != name]
+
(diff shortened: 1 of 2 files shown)


esheldon:
LGTM

esheldon:
Thanks for the PR!

PR #346, opened 2022-01-11
ensure extname is ascii

### LINKED ISSUE #345: The extension name is wrongly sanitized
Hi,

I think I found a bug, and associated fix ;-)

As of today, it is possible to create an extension named `A.B` (which is valid), but the name is parsed as `A_B` at read-time (which is wrong).

In [PyFITSObject_get_hdu_info](https://github.com/esheldon/fitsio/blob/master/fitsio/fitsio_pywrap.c#L694) the `extname` variable -- which I think holds the value of the EXTNAME keyword -- is sanitized by [convert_keyword_to_allowed_ascii](https://github.com/esheldon/fitsio/blob/master/fitsio/fitsio_pywrap.c#L89).

This function converts everything which is not a letter or digit or `-` to `_`. Yet, it seems to me that all ASCII characters are valid as keyword *values*, like the dot, and the function should only be applied to keyword *names*. Or did I completely miss something?

Hope it helps!

esheldon:
closes #345

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -78,16 +78,10 @@ static void convert_to_ascii(char* str) {
 }
 
 /*
-   Replace bad keyword characters with valid keyword ascii characters,
-   namely A-Z,a-z,0-9,_,-
-
-   To make it clear what has happened, the first four characters will be
-   replaced with J U N K and later bad characters with underscore.
-
-   Does not check the keyword is otherwise valid
+   Replace non ascii characters with _
 */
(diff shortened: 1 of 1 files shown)


esheldon:
@kabasset

beckermr:
I may be the culprit here for this bug. Sorry.

kabasset:
@esheldon Thanks a lot for the quick fix!

beckermr (commented):
We may want to edit the function name given it is only used for hdu and ext names now.

beckermr:
Ack I missed that. Sorry!

PR #349, opened 2022-05-17
testCompressPreserveZeros: do not test with the hcompress algorithm

aurel32:
Starting with version 4.1.0, CFITSIO removed the SUBTRACTIVE_DITHER_2
option has been removed when using the HCOMPRESS algorithm. Drop the
corresponding test on the fitsio side so that the testsuite still
passes.


esheldon:
thanks!

PR #361, opened 2023-03-06
ENH remove pkg_resources; update tests

### LINKED ISSUE #360: pkg_resources is deprecated
I'm getting deprecation warnings from importing fitsio lately.  At least on 3.10 (I didn't try other versions).  

It looks like the whole pkg_resources API is deprecated now and fitsio imports it as a top-line import, even though AFAICT it's only used in testing.

The easiest fix would be to move the import into the testGZIPTileCompressedReadLosslessAstropy function, so it would only trigger if someone ran that function.  (Presumably just the CI test suite I suspect.)  But probably better to figure out what the new blessed API is for this and switch over.

To replicate:
```
$ python -W error -c "import fitsio"
  File "<string>", line 1, in <module>
  File "/Users/Mike/mambaforge/envs/py3.10/lib/python3.10/site-packages/fitsio/__init__.py", line 39, in <module>
    from . import test
  File "/Users/Mike/mambaforge/envs/py3.10/lib/python3.10/site-packages/fitsio/test.py", line 7, in <module>
    from pkg_resources import resource_filename
  File "/Users/Mike/mambaforge/envs/py3.10/lib/python3.10/site-packages/pkg_resources/__init__.py", line 121, in <module>
    warnings.warn("pkg_resources is deprecated as an API", DeprecationWarning)
DeprecationWarning: pkg_resources is deprecated as an API
```

beckermr:
Fixes #360

### DIFF
--- a/fitsio/test.py
+++ b/fitsio/test.py
@@ -4,7 +4,13 @@ import tempfile
 import warnings
 import numpy
 from numpy import arange, array
-from pkg_resources import resource_filename
+try:
+    import importlib.resources as importlib_resources
+except ImportError:
+    try:
+        import importlib_resources
+    except ImportError:
+        importlib_resources = None
 import fitsio
 
 from ._fitsio_wrap import cfitsio_use_standard_strings
@@ -1296,9 +1302,13 @@ DATASUM =                      / checksum of the data records\n"""
         """
         Test reading an image gzip compressed by astropy (fixed by cfitsio 3.49)
         """
-        gzip_file = resource_filename(__name__, 'test_images/test_gzip_compressed_image.fits.fz')
-        data = fitsio.read(gzip_file)
-        self.compare_array(data, data*0.0, "astropy lossless compressed image")
+        if importlib_resources is not None:
+            ref = importlib_resources.files(__name__) / 'test_images/test_gzip_compressed_image.fits.fz'
+            with importlib_resources.as_file(ref) as gzip_file:
+                data = fitsio.read(gzip_file)
+            self.compare_array(data, data*0.0, "astropy lossless compressed image")
+        else:
+            self.assertFalse()
 
     def testHCompressTileCompressedWriteRead(self):
         """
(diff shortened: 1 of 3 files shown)


beckermr:
@rmjarvis Care to take a look?

rmjarvis (approved):
Looks fine to me.

FWIW, another option I've seen that might (?) be cleaner for this is pathlib.Path.  Then the pattern is
```
from pathlib import Path
gzip_file = Path(__file__).parent / 'test_images/test_gzip_compressed_image.fits.fz'
```
I think that's fairly portable and not deprecated...  :)  Might be cleaner than the importlib way of doing it.

beckermr:
@esheldon Any comments?

fitsio/test.py
```diff
@@ -1296,8 +1295,19 @@ def testGZIPTileCompressedReadLosslessAstropy(self):
         """
         Test reading an image gzip compressed by astropy (fixed by cfitsio 3.49)
         """
-        gzip_file = resource_filename(__name__, 'test_images/test_gzip_compressed_image.fits.fz')
-        data = fitsio.read(gzip_file)
+        try:
+            import importlib.resources
+            ref = importlib_resources.files("fitsio") / 'test_images' / 'test_gzip_compressed_image.fits.fz'
```
esheldon, line 1300: If we ever want to support windows, we should be using path joining
```
import os
path = os.path.join(
    importlib_resources.files("fitsio"),
    'test_images',
    'test_gzip_compressed_image.fits.fz',
)
```

esheldon (changes requested):

fitsio/test.py
beckermr: This is pathlib. The / is an overloaded python operator, not a character. This is equivalent to os.path.join

esheldon: Thanks, sorry about that. I was mentally thinking of the code Mike posted `gzip_file = Path(__file__).parent / 'test_images/test_gzip_compressed_image.fits.fz'
`

rmjarvis: That works the same way I believe.

rmjarvis: Except it should be `gzip_file = Path(__file__).parent / 'test_images' / 'test_gzip_compressed_image.fits.fz'`

beckermr:
Any other items here @esheldon?

esheldon (approved):

PR #363, opened 2023-03-16
ENH: support for ranges in where()

### LINKED ISSUE #362: Efficient "where" when iterating or slicing
It is currently not practical to use `.where()` with very large tables, because it forces a read of the entire file. Even worse

It would be more convenient if there was a way to apply row filtering when iterating over batches of rows, or table slices more generally. Ideally somewhat like this:

```py
filtered = hdu.filter('<expression>')                  # returns a proxy for the row filter
filtered = hdu['col1', 'col2'].filter('<expression>')  # add row filter to TableColumnSubset
```

This would then simply apply the row filter when indexing the results:

```py
filtered[i:i+n]  # calls fits_find_rows(..., i, n, ...), result has length <= n
```

Indexing into the _filtered_ rows instead might appear more desirable, but the required mapping from filtered to unfiltered row numbers does not seem worth the headache. (*How do you know what `filtered[100_000_000]` is without reading 100 million rows, at least?*)

ntessore:
Add `firstrow` and `lastrow` parameters to the `where()` method of table HDUs to get indices in slices of rows:

```py
fits[1].where('x > 2 && y < 8', 4, 12)  # firstrow, lastrow
```

Closes #362.

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -4542,9 +4542,8 @@ PyFITSObject_where(struct PyFITSObject* self, PyObject* args) {
     int hdutype=0;
     char* expression=NULL;
 
-    LONGLONG nrows=0;
-
-    long firstrow=1;
+    long firstrow;
+    long nrows;
     long ngood=0;
     char* row_status=NULL;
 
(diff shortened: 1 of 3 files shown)


ntessore:
Maybe I should add that I had actually started to implement the

    fits[names][rowfilter][rows]

functionality by adding an `isrowfilter` flag to `_process_args_...` and passing it to all of the `read` methods. But in the end that adds a lot of complexity just to essentially call

    fits[names][where(rowfilter, rows)]

using this very same modified `where()`.

esheldon (changes requested):

PR #365, opened 2023-04-14
fix write_keys() on string cards; add test

CraigLoomis:


esheldon:
Thanks Craig.  I'm surprised this didn't cause anyone else problems.

PR #368, opened 2023-05-25
Return table data in order requested by user

### LINKED ISSUE #367: Return requested row order rather than sorted
Currently we always return rows in sorted order rather than requested order.

E.g. rows=[5, 2, 1]  would return in order [1, 2, 5]

Fix this to return in the users requested order.

also support duplicates in the requested list rather than use unique.  This is not a performance problem because cfitsio uses cacheing

esheldon:
closes #367 

If the users sends `rows = [3, 1]` the old code returned data for `[1, 3]`.  The data are now returned as requested.

Also previously duplicate rows were removed, no longer.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,8 +4,8 @@ version 1.1.9
 Changes
 
     - Row subsets of table data are returned in the order sent by the user
-      rather than sorted and unique.  E.g. rows = [3, 1] will return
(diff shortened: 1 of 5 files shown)


beckermr (approved):
The tests look complete and the C code doesn't have any obvious bugs. LGTM!

PR #378, opened 2023-08-01
ENH hide all symbols

beckermr:
This PR tries to hide all of the cfitsio symbols from the python extension so that the linker does not get confused.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -267,10 +267,8 @@ class build_ext_subclass(build_ext):
                 "could not configure cfitsio %s" % self.cfitsio_version)
 
     def compile_cfitsio(self):
-        import multiprocessing
-        cpus = multiprocessing.cpu_count()
         p = Popen(
-            "make -j %d" % max(cpus, 1),
+            "make",
             shell=True,
             cwd=self.cfitsio_build_dir,
         )
--- a/setup.py
+++ b/setup.py
@@ -269,7 +269,6 @@ class build_ext_subclass(build_ext):
     def compile_cfitsio(self):
         import multiprocessing
         cpus = multiprocessing.cpu_count()
-        assert cpus > 1, cpus
         p = Popen(
             "make -j %d" % max(cpus, 1),
             shell=True,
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -148,7 +148,7 @@ def build_extensions(self):
             # (-fPIC), since we use the python compiler flags
 
             link_objects = glob.glob(
-                os.path.join(self.cfitsio_build_dir, '*.a'))
+                os.path.join(self.cfitsio_build_dir, '*.o'))
```
beckermr, line 151: This is a big change. It is needed to hide the symbols. We now link against the .o files as if we were bundle the C code directly, not just bundling a static lib.

setup.py
```diff
@@ -292,7 +299,11 @@ def check_system_cfitsio_objects(self, obj_name):
 
 sources = ["fitsio/fitsio_pywrap.c"]
 
-ext = Extension("fitsio._fitsio_wrap", sources, include_dirs=['numpy'])
+ext = Extension(
+    "fitsio._fitsio_wrap",
+    sources,
+    include_dirs=['numpy'],
+)
```
beckermr, line 306:
```suggestion
ext = Extension("fitsio._fitsio_wrap", sources, include_dirs=['numpy'])
```

setup.py
```diff
@@ -247,7 +247,9 @@ def copy_update(dir1, dir2):
         args = ''
         if CC is not None:
             args += ' CC="%s"' % ' '.join(CC[:1])
-            args += ' CFLAGS="%s"' % ' '.join(CC[1:])
+            args += ' CFLAGS="%s -fvisibility=hidden"' % ' '.join(CC[1:])
```
esheldon, line 250: Is this needed?  Was using the .o files good enough?

beckermr: We didn't test that with Klaus. I will rebuild and check the libs locally. 

However given the docs on gcc, I think we want both.

beckermr: Actually I think we need it. Before I added the static qualifier to some helper functions in the extension C code, those were appearing as exported symbols in the final lib. I don't think the entire cfitsio code base has static applied.

beckermr: Yup. The symbols are visible without those flags. We did not test explicitly if the thing you are asking about.

esheldon (approved):

PR #381, opened 2023-08-09
Add support for LoongArch.

liuxiang88:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.

Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:
https://loongson.github.io/LoongArch-Documentation/README-EN.html


beckermr (changes requested):
Please do not edit the cfitsio source directly. Instead create a patch file and add it to the set of patches we carry.

PR #390, opened 2024-05-17
Updates for numpy=2.

erykoff:


fitsio/util.py
```diff
@@ -41,7 +41,7 @@ def cfitsio_version(asfloat=False):
             numpy.uint64, numpy.int64)
 
 # different for py3
-_stypes += (numpy.string_, numpy.str_)
+_stypes += (numpy.str_, )
```
beckermr, line 44: We should preserve the use of `numpy.string_` for <2.

beckermr (changes requested):
We need to somehow get numpy2 into the CI to test the builds. Maybe we wait to merge until numpy2 is official on conda-forge.

esheldon:
what does `copy=None` mean in the new numpy?

fitsio/util.py
```diff
@@ -134,9 +134,22 @@ def array_to_native(array, inplace=False):
     return output
 
 
+if numpy.lib.NumpyVersion(numpy.__version__) >= "2.0.0":
+    copy_if_needed = None
```
esheldon, line 138: What does `None` mean?  I don't see that mentioned in the release notes https://numpy.org/devdocs/release/2.0.0-notes.html

erykoff:
I'll restore the `string_` for old numpy on the PR.  The new behavior in numpy 2 is that copy is _always_ `False` as I understand it, and I think they left in `copy=None` as a way to allow backwards compatibility.

fitsio/util.py
erykoff: I'm just following the `scipy` shim code and it does seem to work.

erykoff:
Trying to add py312 and numpy2 to the build matrix.  Let's see what happens!

esheldon:
OK, I guess their release notes are out of date

fitsio/util.py
beckermr: Actually the migration guide says to replace `np.string_` with `np.bytes_` so I think we need that. https://numpy.org/devdocs/numpy_2_0_migration_guide.html#main-namespace

erykoff:
It's a dumb rc thing ...

beckermr:
Yeah copy=None is wrong here. Here are the docs: https://numpy.org/devdocs/release/2.0.0-notes.html#new-copy-keyword-meaning-for-array-and-asarray-constructors

beckermr:
link is busted

```
New copy keyword meaning for array and asarray constructors

Now [numpy.array](https://numpy.org/devdocs/reference/generated/numpy.array.html#numpy.array) and [numpy.asarray](https://numpy.org/devdocs/reference/generated/numpy.asarray.html#numpy.asarray) support three values for copy parameter:

None - A copy will only be made if it is necessary.

True - Always make a copy.

False - Never make a copy. If a copy is required a ValueError is raised.

The meaning of False changed as it now raises an exception if a copy is needed.```

**Comment by beckermr** [comment]
We specify copies or no copies on purpose so properly pass by reference in places iirc.

File: .github/workflows/tests.yml [hunk 3]
```diff
@@ -38,7 +47,7 @@ jobs:
       - uses: conda-incubator/setup-miniconda@v2
         with:
           python-version: ${{ matrix.pyver }}
-          channels: conda-forge,defaults
+          channels: conda-forge,conda-forge/label/numpy_rc,defaults
```
beckermr, line 50: This won't work with strict priority. The rc channel has to come first.

fitsio/util.py
beckermr: See my comments on the main thread. I believe we want to keep False.

erykoff:
So the old numpy1x behavior was `copy=False` would try to do no copy, but would if it needed?  And that is now `copy=None` I think.

fitsio/util.py
erykoff: Then all the tests crash.

beckermr:
Ahhhhhhh yes that's right.

doc string in 1.26:

```
If true (default), then the object is copied. Otherwise, a copy will only be made if __array__ returns a copy, if obj is a nested sequence, or if a copy is needed to satisfy any of the other requirements (dtype, order, etc.).
```

.github/workflows/tests.yml
```diff
@@ -16,15 +16,24 @@ jobs:
...
           - pyver: "3.10"
             npver: "1.20"
+          - pyver: "3.12"
+            npver: "1.20"
+          - pyver: "3.12"
+            npver: "1.21"
+          - pyver: "3.12"
+            npver: "1.23"
+          - pyver: "3.8"
+            npver: "2.0"
```
beckermr, line 35:
```suggestion
            npver: "2.0"
          - pyver: "3.8"
            npver: "1.26"
```

erykoff:
I wonder if all our pushes made the system mad and now won't allocate any nodes for the macos builds ... thoughts?

beckermr:
We probably had too many jobs in the queue and just had to wait for the old ones. We should be able to add bit to the GHA to cancel previous jobs.

erykoff:
All tests have passed and I think this is ready.

beckermr (approved):

PR #394, opened 2024-06-10
fix type warnings with strict gcc

esheldon:


erykoff (approved):

fitsio/fitsio_pywrap.c
```diff
@@ -3603,17 +3626,13 @@ PyFITSObject_read_columns_as_rec(struct PyFITSObject* self, PyObject* args) {
 
     int status=0;
 
-    PyObject* columnsobj=NULL;
-    PyObject* array=NULL;
-
-    PyObject* rowsObj=NULL;
-    PyObject* sortindObj=NULL;
+    PyObject *columns_obj, *array_obj, *rows_obj, *sortind_obj;
```
beckermr, line 3629: Did we want to get rid of the NULL initialization here?

beckermr (commented):
Some of the null inits have been removed. IDK if we want to do this or not?

fitsio/fitsio_pywrap.c
esheldon: They are immediately read into by the module initialization, so its OK in this case. I did it for readability. But if you don't like I'm happy to set NULL

esheldon: by module init I meant things like `PyArg_ParseTuple`

esheldon: Never mind, I'll just set NULL

PR #398, opened 2024-06-12
Cfitsio 4.4.0

### LINKED ISSUE #358: HIERARCH + CONTINUE doesn't work.
fitsio can read/write long value items into a header using CONTINUE lines.  And it can read/write long keyword names with HIERARCH lines.  However, combining the two fails.  Here is a quick unit test, which fails:
```
import fitsio
import numpy as np

header={
    "short1": "Short string",
    "short2": "A value string that is very long.  It needs to be at least 80 characters to trigger using CONTINUE, which this is.",
    "long_keyword_name1": "Another short string",
    "long_keyword_name2": "Another very long string that again must be at least 80 characters.  Probably less this time since it also uses HIERARCH."
}

with fitsio.FITS('temp.fits', 'rw', clobber=True) as f:
    f.write(np.array([1,2,3]), header=header)

with fitsio.FITS('temp.fits') as f:
    header2 = f[0].read_header()

for key in header:
    print(f"{key}\n    {header[key]}\n    {header2[key]}\n")
    assert header[key] == header2[key]
```
Output:
```
short1
    Short string
    Short string

short2
    A value string that is very long.  It needs to be at least 80 characters to trigger using CONTINUE, which this is.
    A value string that is very long.  It needs to be at least 80 characters to trigger using CONTINUE, which this is.

long_keyword_name1
    Another short string
    Another short string

long_keyword_name2
    Another very long string that again must be at least 80 characters.  Probably less this time since it also uses HIERARCH.
    Another very long string that again must be at le

  File "/Users/Mike/GalSim/tmp.py", line 19, in <module>
    assert header[key] == header2[key]
AssertionError
```

Note: This came up on LSST Slack (see thread on software-dev channel Jan 24, 10:28 AM EST).  It seems to be a bug in the backend cfitsio implementation.  Tim Jenness sent an email to cfitsio devs to report the issue.  I'm not sure if it's possible for fitsio to work around this or not to fix the bug sooner, rather than just wait for cfitsio to fix it.

esheldon:
closes #358


esheldon:
@rmjarvis Do you want to test this?

beckermr (approved):

rmjarvis (approved):
Looks like the test you added is sufficient.  I don't think I need to test this separately.  Thanks!

PR #400, opened 2024-06-12
add zlib to manifest

esheldon:


esheldon:
This may address #399

beckermr:
One thing that can be done in ci is we can make an sdist and the try to directly build it.

esheldon:
It makes sense to use test.pypi to test release candidates for this but unfortunately @rainwoodman owns a package called fitsio there.  Is there a way to let me upload to that?

esheldon:
In the mean time I agree we can do the sdist in the CI

esheldon:
The wrong cfitsio version in the manifest was probably the culprit, but including zlib there also makes sense I think.

.github/workflows/tests.yml
```diff
@@ -107,6 +107,23 @@ jobs:
...
+          python setup.py sdist
+          pushd dist/
+          fname=$(ls fitsio*.gz)
+          tar xvfz "$fname"
+          dname=$(echo "$fname" | sed 's/\.tar\.gz//')
+          pushd $dname
+          pip install -vv -e .
+          SKIP_BZIP_TEST=true pytest -vv fitsio
+          popd
+          popd
```
beckermr, line 125:
```suggestion
          pip install -vv --no-deps --no-build-isolation dist/*.tar.gz
```

This version does not produce an editable install, but you can install the sdist directly without unpacking it. It might be worth check that the install itself doesn't fail this way even if we cannot run the tests.

esheldon: How would you  run tests?

esheldon:
I don't have a way to test this with test.pypi so I'm going to try for a release at 1.2.3 unless there is an objection

.github/workflows/tests.yml
esheldon: Are you suggesting we add another section to the CI which just runs that command but does not do tests?

beckermr: yes

beckermr: you can try something like this

```
          rm -rf dist/*
          python setup.py sdist
          pip install -vv --no-deps --no-build-isolation dist/*.tar.gz
          cd ..
          python -c "import conda_forge_tick; assert conda_forge_tick.__version__ != '0.0.0'"
          cd -
          pip uninstall conda-forge-tick --yes
```

esheldon: `ModuleNotFoundError: No module named 'conda_forge_tick'`

.github/workflows/tests.yml
```diff
@@ -107,20 +107,41 @@ jobs:
...
+          find . -name "*.so" -type f -delete
+          rm -rf dist
+
+          python setup.py sdist
+
+          pip install -vv --no-deps --no-build-isolation dist/*.tar.gz
+          cd ..
+          python -c "import conda_forge_tick; assert conda_forge_tick.__version__ != '0.0.0'"
+          cd -
+          pip uninstall conda-forge-tick --yes
```
beckermr, line 124:
```suggestion
          python -c "import fitsio; assert fitsio.__version__ != '0.0.0'"
          cd -
          pip uninstall fitsio --yes
```

esheldon: OK, that makes a lot more sense

.github/workflows/tests.yml
beckermr: yeah you need `fitsio` there. ;p

PR #411, opened 2024-08-16
Seed dithering

### LINKED ISSUE #351: Feature request: allow setting the FPACK dither seed
Hi,

I'm trying to make it so my code writes bit-identical outputs every time it runs, but I'm using the compression schemes that do the subtractive dither thing, so there is a seed chosen at run-time.  It looks like CFITSIO has routines to set that seed -
`fits_set_quantize_dither()` or `fits_set_dither_seed()` etc
https://heasarc.gsfc.nasa.gov/docs/software/fitsio/c/c_user/node41.html

It looks like one place to add this feature would be here,
https://github.com/esheldon/fitsio/blob/c8f8d608efbbc304ff58c9e005a1f382b61abfef/fitsio/fitsio_pywrap.c#L1376
ie similar to how the quantization level can be set,
https://github.com/esheldon/fitsio/blob/c8f8d608efbbc304ff58c9e005a1f382b61abfef/fitsio/fitsio_pywrap.c#L1483

thanks!
-dustin

esheldon:
closes #351

Add ability to seed the subtractive dither

### DIFF
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -1846,21 +1846,40 @@ def get_qmethod(qmethod):
 
 
 def get_dither_seed(dither_seed):
-    if dither_seed is None or dither_seed in ['clock', 'CLOCK']:
-        # 0 means do not set the seed, use the system clock. This is the
-        # default
+    """
+    Convert a seed value or indicator to the approprate integer value for
+    cfitsio
+
+    Parameters
+    ----------
+    dither_seed: number or string
+        Seed for the subtractive dither.  Seeding makes the lossy compression
+        reproducible.  Allowed values are
+            None or 0 or 'clock':
+                Return 0, do not set the seed explicitly, use the system clock
+            negative or 'checksum':
+                Return -1, means Set the seed based on the data checksum
+            1-10_000:
+                use the input seed
+    """
+    if isinstance(dither_seed, (str, bytes)):
+        dlow = dither_seed.lower()
+        if dlow == 'clock':
+            seed_out = 0
+        elif dlow == 'checksum':
+            seed_out = -1
+        else:
+            raise ValueError(f'Bad dither_seed {dither_seed}')
+    elif dither_seed is None:
         seed_out = 0
-    elif dither_seed in ['checksum', 'CHECKSUM']:
-        seed_out = -1
     else:
         # must fit in an int
-        seed_out = numpy.int32(dither_seed)
+        seed_out = numpy.int32(dither_seed).clip(min=-1)
 
-        if seed_out < -1 or seed_out > 10_000:
-            raise ValueError(
-                f'Got dither_seed {seed_out}, expected -1, 0 or '
-                'a seed in range [1, 10000]'
-            )
+    if seed_out > 10_000:
+        raise ValueError(
+            f'Got dither_seed {seed_out}, expected avalue <= 10_000'
+        )
 
     return seed_out
 
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -1862,7 +1862,10 @@ def get_dither_seed(dither_seed):
             1-10_000:
                 use the input seed
     """
-    if isinstance(dither_seed, (str, bytes)):
+    if isinstance(dither_seed, bytes):
+        dither_seed = str(dither_seed, 'utf-8')
+
+    if isinstance(dither_seed, str):
         dlow = dither_seed.lower()
         if dlow == 'clock':
             seed_out = 0
--- a/fitsio/fitslib.py
+++ b/fitsio/fitslib.py
@@ -377,9 +377,16 @@ def write(filename, data, extname=None, extver=None, header=None,
                 Preserves zeros
 
         Defaults to 'SUBTRACTIVE_DITHER_1' which follows the fpack defaults
-    dither_seed: int
-        Seed for the subtractive dither.  Seeding makes the lossy
-        compression reproducible.
+
+    dither_seed: int or None
+        Seed for the subtractive dither.  Seeding makes the lossy compression
+        reproducible.  Allowed values are
+            None or 0 or 'clock':
+                do not set the seed explicitly, use the system clock
+            -1 or 'checksum':
+                Set the seed based on the data checksum
+            1-10_000:
+                use the input seed
 
     hcomp_scale: float
         Scale value for HCOMPRESS, 0.0 means lossless compression. Default is
@@ -692,9 +699,15 @@ class FITS(object):
                     Preserves zeros
 
             Defaults to 'SUBTRACTIVE_DITHER_1' which follows the fpack defaults
-        dither_seed: int
+        dither_seed: int or None
             Seed for the subtractive dither.  Seeding makes the lossy
-            compression reproducible.
+            compression reproducible.  Allowed values are
+                None or 0 or 'clock':
+                    do not set the seed explicitly, use the system clock
+                -1 or 'checksum':
+                    Set the seed based on the data checksum
+                1-10_000:
+                    use the input seed
 
         hcomp_scale: float
             Scale value for HCOMPRESS, 0.0 means lossless compression. Default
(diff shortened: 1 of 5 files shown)


esheldon:
@dstndstn does this work for you?

beckermr (changes requested):
Are there other parts of the lossy compression seed setting in the fits standard / cfitsio that we should be aware of to make sure our APIs here can support? We need not implement them OFC.

esheldon:
Possibly.  This code I added only sets the seed when you create the HDU, which seems safe.  What other operations should be supported?

beckermr:
The other options allowed by the standard are

```
-qt <level> - compute the seed from the checksum of the first tile of image pixels
-qN <level> where N is an integer between 1 and 10000 – use seed N (e.g., -q3010 4).
```

This all gets coupled into the FITS header via

```
FZDTHRSD - 'CLOCK', 'CHECKSUM', 1 - 10000
```

and here is the doc string for `fits_set_dither_seed`

```
This routine specifies the value of the offset that should be applied when calculating the random dithering when quantizing floating point iamges. A random offset should be applied to each image to avoid quantization effects when taking the difference of 2 images, or co-adding a set of images. Without this random offset, the corresponding pixel in every image will have exactly the same dithering.

offset = 0 means use the default random dithering based on system time offset = negative means randomly chose dithering based on 1st tile checksum offset = [1 - 10000] means use that particular dithering pattern

```

The default according to the fpackguide is 

```
By default, the initial seed value is randomly chosen based on the
current system clock time.
```

So we could support the following options for `dither_seed` at the python level:

- 0 or 'CLOCK' or 'clock': set seed to 0
- negative number or 'CHECKSUM' or 'checksum': set seed to -1
- 1-10000: send seed through
- error for anything else

esheldon:
yikes, I should have red the docs first

fitsio/fitslib.py
```diff
@@ -1800,6 +1845,26 @@ def get_qmethod(qmethod):
     return _qmethod_map[qmethod]
 
 
+def get_dither_seed(dither_seed):
+    if dither_seed is None or dither_seed in ['clock', 'CLOCK']:
```
beckermr, line 1849:
```suggestion
    if dither_seed is None or (isinstance(dither_seed, str) and dither_seed.lower() == 'clock'):
```

fitsio/fitslib.py
```diff
@@ -1800,6 +1845,26 @@ def get_qmethod(qmethod):
...
+        # 0 means do not set the seed, use the system clock. This is the
+        # default
+        seed_out = 0
+    elif dither_seed in ['checksum', 'CHECKSUM']:
```
beckermr, line 1853:
```suggestion
    elif isinstance(dither_seed, str) and dither_seed.lower() == 'checksum':
```

fitsio/fitslib.py
```diff
@@ -1800,6 +1845,26 @@ def get_qmethod(qmethod):
...
+        seed_out = -1
+    else:
+        # must fit in an int
+        seed_out = numpy.int32(dither_seed)
+
+        if seed_out < -1 or seed_out > 10_000:
```
beckermr, line 1859:
```suggestion
        if seed_out > 10_000:
```

Any negative value is allowed by the fits routine. Fun!

fitsio/fitslib.py
```diff
@@ -1800,6 +1845,26 @@ def get_qmethod(qmethod):
...
+            raise ValueError(
+                f'Got dither_seed {seed_out}, expected -1, 0 or '
+                'a seed in range [1, 10000]'
```
beckermr, line 1862:
```suggestion
                f'Got dither_seed {seed_out}, expected a value < 10000'
```

beckermr (changes requested):
A few tweaks but looking good. We'll need to adjust the tests to cover more cases and not error on -3. We should also test that -3 and -1 give the same answer.

fitsio/fitslib.py
esheldon: These are equivalent, and I actually think the first reads better

esheldon: actually no, I guess you are suggesting we accept things like `CloCk`

beckermr: I agree it is an odd edge case, but I can imagine someone putting in a single capitol by accident and getting super frustrated why things are broken / not working.

beckermr (approved):
LGTM! Thanks!

fitsio/fitslib.py
```diff
@@ -377,6 +378,16 @@ def write(filename, data, extname=None, extver=None, header=None,
 
         Defaults to 'SUBTRACTIVE_DITHER_1' which follows the fpack defaults
 
+    dither_seed: int or None
+        Seed for the subtractive dither.  Seeding makes the lossy compression
+        reproducible.  Allowed values are
+            None or 0 or 'clock':
+                do not set the seed explicitly, use the system clock
+            -1 or 'checksum':
```
beckermr, line 386:
```suggestion
            negative or 'checksum':
```

esheldon:
OK, we'll see what Dustin says

fitsio/fitslib.py
```diff
@@ -686,6 +699,15 @@ def write(self, data, units=None, extname=None, extver=None,
                     Preserves zeros
 
             Defaults to 'SUBTRACTIVE_DITHER_1' which follows the fpack defaults
+        dither_seed: int or None
+            Seed for the subtractive dither.  Seeding makes the lossy
+            compression reproducible.  Allowed values are
+                None or 0 or 'clock':
+                    do not set the seed explicitly, use the system clock
+                -1 or 'checksum':
```
beckermr, line 707:
```suggestion
                negative or 'checksum':
```

fitsio/fitslib.py
```diff
@@ -792,6 +816,15 @@ def write_image(self, img, extname=None, extver=None,
...
```
beckermr, line 824:
```suggestion
                negative or 'checksum':
```

fitsio/fitslib.py
```diff
@@ -917,7 +952,15 @@ def create_image_hdu(self,
                     Preserves zeros
 
             Defaults to 'SUBTRACTIVE_DITHER_1' which follows the fpack defaults
-
+        dither_seed: int or None
+            Seed for the subtractive dither.  Seeding makes the lossy
+            compression reproducible.  Allowed values are
+                None or 0 or 'clock':
+                    do not set the seed explicitly, use the system clock
+                -1 or 'checksum':
```
beckermr, line 960:
```suggestion
                negative or 'checksum':
```

fitsio/fitslib.py
```diff
@@ -1800,6 +1845,45 @@ def get_qmethod(qmethod):
...
+            seed_out = 0
+        elif dlow == 'checksum':
+            seed_out = -1
+        else:
+            raise ValueError(f'Bad dither_seed {dither_seed}')
+    elif dither_seed is None:
+        seed_out = 0
+    else:
+        # must fit in an int
+        seed_out = numpy.int32(dither_seed).clip(min=-1)
```
beckermr, line 1877: Why clip here?

esheldon: that's leftover from when I thought -1 was the lowest allowed

fitsio/fitslib.py
```diff
@@ -1800,6 +1845,45 @@ def get_qmethod(qmethod):
...
+        Seed for the subtractive dither.  Seeding makes the lossy compression
+        reproducible.  Allowed values are
+            None or 0 or 'clock':
+                Return 0, do not set the seed explicitly, use the system clock
+            negative or 'checksum':
+                Return -1, means Set the seed based on the data checksum
+            1-10_000:
+                use the input seed
+    """
+    if isinstance(dither_seed, (str, bytes)):
```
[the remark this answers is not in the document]
beckermr: Also this behavior is utterly baffling to me:

```
In [7]: b = b"Clock"

In [8]: b
Out[8]: b'Clock'

In [9]: type(b)
Out[9]: bytes

In [10]: b.lower()
Out[10]: b'clock'
```

If we are supposed to think of bytes as just bytes and we are forced to explicitly decode them to a string, then a method like `lower()` doesn't make sense IMHO.

beckermr (approved):

PR #422, opened 2025-02-21
When using file-level compression, allow setting compression parameters

dstndstn:
In the `legacypipe` code, we use `mem://` files very extensively -- this allows us to do in-memory checksums before writing to disk.

We also use compression extensively, which we set at the file level via extended filename syntax - eg we create files like
```
fits = fitsio.FITS('mem://[compress R 100,100; qz -1e-4]', 'rw')
```

However, we find that when doing this, when you go to write an image HDU, the compression parameters don't get passed through because `compress=` is not set.

This patch fixes that.

Test case: these two files should get the same `ZDITHER0` set.
```
import numpy as np
import fitsio
A = ((1+np.arange(10000)) * 0.1).astype(np.float32)

fitsio.write('dither1.fits', A.copy(), dither_seed='checksum', compress='RICE', qlevel=1e-4,
             tile_dims=100, clobber=True)
hdr = fitsio.read_header('dither1.fits', ext=1)
print('File #1: ZDITHER0:', hdr['ZDITHER0'])

fits = fitsio.FITS('mem://[compress R 100,100; qz -1e-4]', 'rw')
fits.write(A.copy(), dither_seed='checksum')
data = fits.read_raw()
fits.close()
f = open('dither2.fits', 'wb')
f.write(data)
f.close()
hdr = fitsio.read_header('dither2.fits', ext=1)
print('File #2: ZDITHER0:', hdr['ZDITHER0'])
```

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -1482,6 +1482,10 @@ PyFITSObject_create_image_hdu(struct PyFITSObject* self, PyObject* args, PyObjec
             write_data=1;
         }
 
+	if (comptype == 0) {
+	    comptype = (*(self->fits)->Fptr).request_compress_type;
+	}
+
         // 0 means NOCOMPRESS but that wasn't defined in the bundled version of cfitsio
         // if (comptype >= 0) {
         if (comptype > 0) {
--- a/fitsio/tests/test_image_compression.py
+++ b/fitsio/tests/test_image_compression.py
@@ -347,6 +347,7 @@ def test_compressed_seed_bad(dither_seed):
                 dither_seed=dither_seed,
             )
 
+
 def test_memory_compressed_seed():
     import fitsio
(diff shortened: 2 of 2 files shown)


esheldon:
Thanks Dustin.  Can you add a unit test for this one?

fitsio/fitsio_pywrap.c
```diff
@@ -1482,6 +1482,10 @@ PyFITSObject_create_image_hdu(struct PyFITSObject* self, PyObject* args, PyObjec
             write_data=1;
         }
 
+	if (comptype == 0) {
+	    comptype = (*(self->fits)->Fptr).request_compress_type;
+	}
```
beckermr, line 1487:
```suggestion
        if (comptype == 0) {
            comptype = (*(self->fits)->Fptr).request_compress_type;
        }
```

PR #427, opened 2025-05-14
Make sure string column with vec size 1 work

### LINKED ISSUE #426: Crash when trying to write a table with a string array of length 1
```
import numpy as np
import fitsio

data = np.zeros(10, dtype=[("string", ("U10", 1))])
fitsio.write("test.fits", data)
```
This crashes with:
```
File ~/lsst/20250424/conda/envs/lsst-scipipe-10.0.0/lib/python3.12/site-packages/fitsio/fitslib.py:409, in write(filename, data, extname, extver, header, clobber, ignore_empty, units, table_type, names, write_bitcols, compress, tile_dims, qlevel, qmethod, dither_seed, hcomp_scale, hcomp_smooth, **keys)
    404 kwargs = {
    405     'clobber': clobber,
    406     'ignore_empty': ignore_empty
    407 }
    408 with FITS(filename, 'rw', **kwargs) as fits:
--> 409     fits.write(
    410         data,
    411         table_type=table_type,
    412         units=units,
    413         extname=extname,
    414         extver=extver,
    415         header=header,
    416         names=names,
    417         write_bitcols=write_bitcols,
    418 
    419         compress=compress,
    420         tile_dims=tile_dims,
    421         qlevel=qlevel,
    422         qmethod=qmethod,
    423         dither_seed=dither_seed,
    424         hcomp_scale=hcomp_scale,
    425         hcomp_smooth=hcomp_smooth,
    426     )

File ~/lsst/20250424/conda/envs/lsst-scipipe-10.0.0/lib/python3.12/site-packages/fitsio/fitslib.py:761, in FITS.write(self, data, units, extname, extver, compress, tile_dims, qlevel, qmethod, dither_seed, hcomp_scale, hcomp_smooth, header, names, table_type, write_bitcols, **keys)
    751     self.write_image(data, extname=extname, extver=extver,
    752                      compress=compress,
    753                      tile_dims=tile_dims,
   (...)    758                      hcomp_smooth=hcomp_smooth,
    759                      header=header)
    760 else:
--> 761     self.write_table(data, units=units,
    762                      extname=extname, extver=extver, header=header,
    763                      names=names,
    764                      table_type=table_type,
    765                      write_bitcols=write_bitcols)

File ~/lsst/20250424/conda/envs/lsst-scipipe-10.0.0/lib/python3.12/site-packages/fitsio/fitslib.py:1182, in FITS.write_table(self, data, table_type, names, formats, units, extname, extver, header, write_bitcols)
   1179     self[-1].write_keys(header)
   1180     self[-1]._update_info()
-> 1182 self[-1].write(data, names=names)

File ~/lsst/20250424/conda/envs/lsst-scipipe-10.0.0/lib/python3.12/site-packages/fitsio/hdu/table.py:291, in TableHDU.write(self, data, firstrow, columns, names, slow, **keys)
    288         nonobj_arrays.append(colref)
    290 for tcolnum, tdata in zip(nonobj_colnums, nonobj_arrays):
--> 291     self._verify_column_data(tcolnum, tdata)
    293 if len(nonobj_arrays) > 0:
    294     self._FITS.write_columns(
    295         self._ext+1, nonobj_colnums, nonobj_arrays,
    296         firstrow=firstrow+1, write_bitcols=self.write_bitcols)

File ~/lsst/20250424/conda/envs/lsst-scipipe-10.0.0/lib/python3.12/site-packages/fitsio/hdu/table.py:417, in TableHDU._verify_column_data(self, colnum, data)
    411     raise ValueError(
    412         "bad input data for column '%s': "
    413         "expected '%s', got '%s'" % (
    414             col_name, npy_type, this_npy_type))
    416 if this_shape != col_shape:
--> 417     raise ValueError(
    418         "bad input shape for column '%s': "
    419         "expected '%s', got '%s'" % (col_name, col_shape, this_shape))

ValueError: bad input shape for column 'string': expected '(1,)', got '()'
```

This works when the string array has length 2 or greater.

esheldon:
closes #426 

This fixes a bug where the shape of the column was not correct.

Adds tests that string and number vec1 column data get returned, even though read as scalar

### DIFF
--- a/fitsio/hdu/table.py
+++ b/fitsio/hdu/table.py
@@ -368,17 +368,18 @@ class TableHDU(HDUBase):
         verify the input data is of the correct type and shape
         """
         this_dt = data.dtype.descr[0]
+        npy_type, isvar, istbit = self._get_tbl_numpy_dtype(colnum)
+        is_string = npy_type[0] in ('S', 'U')
 
         if len(data.shape) > 2:
             this_shape = data.shape[1:]
-        elif len(data.shape) == 2 and data.shape[1] > 1:
+        elif len(data.shape) == 2 and (data.shape[1] > 1 or is_string):
             this_shape = data.shape[1:]
         else:
             this_shape = ()
 
         this_npy_type = this_dt[1][1:]
 
-        npy_type, isvar, istbit = self._get_tbl_numpy_dtype(colnum)
         info = self._info['colinfo'][colnum]
 
         if npy_type[0] in ['>', '<', '|']:
@@ -387,7 +388,9 @@ class TableHDU(HDUBase):
         col_name = info['name']
         col_tdim = info['tdim']
         col_shape = _tdim2shape(
-            col_tdim, col_name, is_string=(npy_type[0] == 'S'))
+            col_tdim, col_name,
+            is_string=is_string
+        )
 
         if col_shape is None:
             if this_shape == ():
@@ -1267,7 +1270,7 @@ class TableHDU(HDUBase):
         shape = None
         tdim = info['tdim']
 
-        shape = _tdim2shape(tdim, name, is_string=(npy_type[0] == 'S'))
+        shape = _tdim2shape(tdim, name, is_string=(npy_type[0] in ('S', 'U')))
         if shape is not None:
             if nrows > 1:
                 if not isinstance(shape, tuple):
@@ -1333,7 +1336,8 @@ class TableHDU(HDUBase):
             tdim = self._info['colinfo'][colnum]['tdim']
             shape = _tdim2shape(
                 tdim, name,
-                is_string=(npy_type[0] == 'S' or npy_type[0] == 'U'))
+                is_string=(npy_type[0] in ('S', 'U'))
+            )
             if shape is not None:
                 descr = (name, npy_type, shape)
             else:
--- a/fitsio/tests/test_table.py
+++ b/fitsio/tests/test_table.py
@@ -86,6 +86,108 @@ def test_table_read_write():
                         )
 
 
+@pytest.mark.parametrize('nvec', [2, 1])
+def test_table_read_write_vec1(nvec):
+    """
+    ensure the data for vec length 1 gets round-tripped, even though
+    the shape is not preserved
+    """
+    dtype = [('x', 'f4', (nvec,))]
+    num = 10
+    data = np.zeros(num, dtype=dtype)
+    data['x'] = np.arange(num * nvec).reshape(num, nvec)
+    assert data['x'].shape == (num, nvec)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        fname = os.path.join(tmpdir, 'test.fits')
+
+        with FITS(fname, 'rw') as fits:
+            fits.write_table(data)
+
+            d = fits[1].read()
+            if nvec == 1:
+                assert d['x'].shape == (num,)
+            compare_array(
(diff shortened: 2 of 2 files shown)


erykoff (approved):
Thanks!

fitsio/hdu/table.py
```diff
@@ -368,17 +368,18 @@ def _verify_column_data(self, colnum, data):
         verify the input data is of the correct type and shape
         """
         this_dt = data.dtype.descr[0]
+        npy_type, isvar, istbit = self._get_tbl_numpy_dtype(colnum)
+        is_string = npy_type[0] in ('S', 'U')
 
         if len(data.shape) > 2:
             this_shape = data.shape[1:]
-        elif len(data.shape) == 2 and data.shape[1] > 1:
+        elif len(data.shape) == 2 and (data.shape[1] > 1 or is_string):
```
beckermr, line 376: Can you add a comment here on why strings are a special case?

fitsio/tests/test_table.py
```diff
@@ -86,6 +91,130 @@ def test_table_read_write():
...
+                    "table single field read 'string'"
+                )
+            else:
+                if nvec == 1:
+                    assert d['string'].shape == (num,)
+
+                compare_array(
+                    data['string'].ravel(), d['string'].ravel(),
+                    "table single field read 'string'"
+                )
```
beckermr, line 177: Why is there a branch here? We should add a comment in the code.

beckermr (changes requested):
The code looks fine but I think I am missing some details on what the special case is exactly. I've asked for a few comments in places to help future us understand this edge case.

beckermr (changes requested):
I missed that this is a breaking change wrt the behavior of returning dims of (1,) versus a scalar. We should figure that bit out.

See also https://github.com/esheldon/fitsio/issues/426#issuecomment-2881224081

esheldon:
It is a breaking change, in that those columns will now be properly round tripped whereas they were not before.

On the other hand, I consider this a bug in numpy that I couldn't get this to work in 1.0.  Seen that way it is breaking in the sense of correcting behavior.

beckermr:
Is there a way we can make it work for non-strings?

esheldon:
There is no way to set the TDIM for length-1 vector columns, they must be length 2 or more.

I can see hacks working.  We could add a header keyword that tells us the situation for a given column so fitsio can reshape the data.

beckermr:
My feeling on this specific item is that this specific behavior of FITSIO is so ingrained in people's codes + minds that if we break it, we're going to cause silent or at least very weird bugs. I get that it can be considered a bug, but long-standing bugs that people code around effectively become part of the API.

beckermr:
> I can see hacks working. We could add a header keyword that tells us the situation for a given column so fitsio can reshape the data.

Right. My feeling is that we

1. Keep the current behavior before this PR where columns of shape `(1,)` are cast to scalars for all times.

2. If we do want to fix this, we fix it for all types AND we bump fitsio to major version 2.

This seems to me to be the only way to do this without causing lots of bad side effects.

esheldon:
I don't actually know how to workaround the new numpy behavior, will have to figure it out.

beckermr:
Can't we just change the output dtype for strings if we find TDIM has 2 elements?

esheldon:
There is code reuse for read and write.  It seems like different behavior might be needed for read vs write in this case, but need to look into it.

PR #435, opened 2025-08-10
Fix potential free of null pointer in stringlist_delete

### LINKED ISSUE #433: Perform null check before free
https://github.com/esheldon/fitsio/blob/90697250a609cee86a9b3c4d9567bd0fa1da8260/fitsio/fitsio_pywrap.c#L303-L315

Line 308, probably want to null check before attempting to free.

Also the `free(slist->data);` should be inside the conditional block.

cruzzil:
Closes #433


esheldon:
If the pointer given to free is null, the function does nothing

PR #436, opened 2025-08-16
fix: ensure arrays are aligned before calling fitsio functions

### LINKED ISSUE #434: Ensure Numpy arrays are aligned
Should ensure that the numpy arrays are aligned to the dtype before passing them to cfitsio. On some architectures this is undefined behaviour or a crash.

E.g.

```
def array_to_native_c(array_in, inplace=False):
    # copy only made if not C order
    # arr = numpy.array(array_in, order="C", copy=copy_if_needed)
    arr = numpy.require(array_in, requirements=['C_CONTIGUOUS', 'ALIGNED'])
    return array_to_native(arr, inplace=inplace)
```

Need to do similar in `array_to_native` function as well.

petesmc:
Closes #434


beckermr:
Here is the relevant text from the docs:

> There are 4 relevant uses of the word align used in NumPy:
> 
> The [dtype.alignment](https://numpy.org/doc/stable/reference/generated/numpy.dtype.alignment.html#numpy.dtype.alignment) attribute (descr->alignment in C). This is meant to reflect the “true alignment” of the type. It has arch-dependent default values for all datatypes, except for the structured types created with align=True as described below.
> 
> The ALIGNED flag of an ndarray, computed in IsAligned and checked by [PyArray_ISALIGNED](https://numpy.org/doc/stable/reference/c-api/array.html#c.PyArray_ISALIGNED). This is computed from [dtype.alignment](https://numpy.org/doc/stable/reference/generated/numpy.dtype.alignment.html#numpy.dtype.alignment). It is set to True if every item in the array is at a memory location consistent with [dtype.alignment](https://numpy.org/doc/stable/reference/generated/numpy.dtype.alignment.html#numpy.dtype.alignment), which is the case if the data ptr and all strides of the array are multiples of that alignment.
> 
> The align keyword of the dtype constructor, which only affects [Structured arrays](https://numpy.org/doc/stable/user/basics.rec.html#structured-arrays). If the structure’s field offsets are not manually provided, NumPy determines offsets automatically. In that case, align=True pads the structure so that each field is “true” aligned in memory and sets [dtype.alignment](https://numpy.org/doc/stable/reference/generated/numpy.dtype.alignment.html#numpy.dtype.alignment) to be the largest of the field “true” alignments. This is like what C-structs usually do. Otherwise if offsets or itemsize were manually provided align=True simply checks that all the fields are “true” aligned and that the total itemsize is a multiple of the largest field alignment. In either case [dtype.isalignedstruct](https://numpy.org/doc/stable/reference/generated/numpy.dtype.isalignedstruct.html#numpy.dtype.isalignedstruct) is also set to True.
> 
> IsUintAligned is used to determine if an ndarray is “uint aligned” in an analogous way to how IsAligned checks for true alignment.

Apparently the `align` keyword specifically controls how structured arrays are laid out in memory: https://numpy.org/doc/stable/reference/generated/numpy.dtype.html#numpy.dtype

esheldon:
Yes, align to dtype controls it, adding extra padding one would expect for a C struct, but this is *not* what the align flag means

esheldon:
(see my example above to see how `align=True` to dtype constructor, and align flag don't mean the same thing)

beckermr:
Yep!

I do not think the numpy-specifc getters and setters are always enough. 

If we ever pass the memory location of a numpy array down into cfitsio, then cfitsio will dereference it as a pointer to the corresponding C type directly. In that case, the array has to be aligned or the deferencing mayt cause a crash, at least according to the numpy docs. They say

> Even if you have a [contiguous](https://numpy.org/doc/stable/glossary.html#term-contiguous) chunk of memory, you cannot just assume it is safe to dereference a datatype-specific pointer to an element. Only if the [NPY_ARRAY_ALIGNED](https://numpy.org/doc/stable/reference/c-api/array.html#c.NPY_ARRAY_ALIGNED) flag is set, this is a safe operation. On some platforms it will work but on others, like Solaris, it will cause a bus error.

So I think this PR is a no-op in most cases, but on some platforms is needed to avoid crashes when passing references to numpy-managed memory directly into cfitsio.

beckermr:
Yeah the original issue concerned a port of cfitsio to rust, so maybe it isn't surprising you can't reproduce locally.

beckermr:
Google's AI said to do this:

import numpy as np

# Create a base aligned array
aligned_array = np.arange(10, dtype=np.int32)

# Create a view that starts at an offset, making it "unaligned" relative to its itemsize
# For example, shifting by 1 byte (assuming int32, which is 4 bytes)
# This will likely cause issues if you try to perform operations on it directly.
unaligned_view = np.ndarray(
    shape=aligned_array.shape,
    dtype=aligned_array.dtype,
    buffer=aligned_array.data,
    offset=1, # Offset by 1 byte
    strides=aligned_array.strides
)

print(f"Aligned array data pointer: {aligned_array.ctypes.data}")
print(f"Unaligned view data pointer: {unaligned_view.ctypes.data}")

esheldon:
Gives a `ValueError`

beckermr:
AI was wrong?! lol worth a try.

fitsio/util.py
```diff
@@ -152,7 +154,8 @@ def array_to_native(array, inplace=False):
 
 def array_to_native_c(array_in, inplace=False):
     # copy only made if not C order
-    arr = numpy.array(array_in, order='C', copy=copy_if_needed)
+    # arr = numpy.array(array_in, order='C', copy=copy_if_needed)
+    arr = numpy.require(array_in, requirements=['C_CONTIGUOUS', 'ALIGNED'])
```
beckermr, line 158:
```suggestion
    arr = numpy.require(
        array_in,
        requirements=['C_CONTIGUOUS', 'ALIGNED'],
    )
```

beckermr:
I also think adding `-fsanitize=alignment` to our compiler flags when testing would be helpful.

beckermr:
How do you propose we detect platforms that require alignment? I have no idea how to do that.

.github/workflows/tests.yml
```diff
@@ -71,6 +74,7 @@ jobs:
       - name: test non-bundled build
         shell: bash -l {0}
         run: |
+          export CFLAGS="${CFLAGS} -fno-sanitize-recover=alignment"
```
beckermr, line 77:
```suggestion
          export CFLAGS="${CFLAGS} ${TEST_CFLAGS}"
```

fitsio/util.py
```diff
@@ -152,7 +154,10 @@ def array_to_native(array, inplace=False):
 
 def array_to_native_c(array_in, inplace=False):
     # copy only made if not C order
-    arr = numpy.array(array_in, order='C', copy=copy_if_needed)
+    arr = numpy.require(
+        array_in,
+        requirements=['C_CONTIGUOUS', 'ALIGNED'],
```
beckermr, line 159:
```suggestion
        requirements=['C_CONTIGUOUS',],  # 'ALIGNED'],
```

fitsio/util.py
```diff
@@ -134,7 +134,9 @@ def array_to_native(array, inplace=False):
     else:
         output = array
 
-    return output
+    output_aligned = numpy.require(output, requirements=['ALIGNED'])
```
beckermr, line 137:
```suggestion
    return output
    # output_aligned = numpy.require(output, requirements=['ALIGNED'])
    
```

fitsio/util.py
```diff
@@ -135,6 +135,10 @@ def array_to_native(array, inplace=False):
         output = array
 
     return output
+    # output_aligned = numpy.require(output, requirements=['ALIGNED'])
+    
+
+    return output_aligned
```
beckermr, line 141:
```suggestion
    return numpy.require(output, requirements=['ALIGNED'])
```

fitsio/util.py
```diff
@@ -152,7 +156,10 @@ def array_to_native(array, inplace=False):
 
 def array_to_native_c(array_in, inplace=False):
     # copy only made if not C order
-    arr = numpy.array(array_in, order='C', copy=copy_if_needed)
+    arr = numpy.require(
+        array_in,
+        requirements=['C_CONTIGUOUS',],  # 'ALIGNED'],
```
beckermr, line 161:
```suggestion
        requirements=['C_CONTIGUOUS', 'ALIGNED'],
```

.github/workflows/tests.yml
```diff
@@ -8,6 +8,8 @@ on:
 
 env:
   PY_COLORS: "1"
+  TEST_CFLAGS: "-fsanitize=alignment -fno-sanitize-recover=alignment"
```
beckermr, line 11:
```suggestion
  # These compiler flags force the tests to fail if arrays are
  # accessed at the C level from an unaligned location.
  TEST_CFLAGS: "-fsanitize=alignment -fno-sanitize-recover=alignment"
```

PR #441, opened 2025-09-04
feat: move to setuptools-scm version

beckermr:
In the version bump to 1.2.7, we missed properly setting the version in the `setup.py` and in the package. Between the two of us, we have messed this up more than once. This PR moves the repo to setuptools-scm which pulls the version from the tag. It should ensure we don't mess this up anymore.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,8 +16,6 @@ import glob
 import shutil
 
 
-print(sys.argv, flush=True)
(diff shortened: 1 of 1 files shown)


beckermr:
Yes. The sdist and the GitHub release tarball will both have the right version.

PR #443, opened 2025-09-15
fix: pypi uploads broke after setuptools-scm

beckermr:
There were some missing refs for the checkout and the upload event did not recognize workflow dispatches.


beckermr:
I am going to merge this one with the final pypi upload step disabled to ensure the wheels get the right version. Then I will turn it back on to push the 1.2.7 release to pypi.

PR #455, opened 2025-10-17
fix: ensure we properly write into image slices

### LINKED ISSUE #284: Overwrite subset of image
Hi,

I would like to overwrite a subset of an existing image so that the new, smaller, image appears as a rectangle of new values. Instead, I find that the new values appear in rows wrapped around the original image, starting at the place of the 'start' pixel.

Example code:

  ```img=np.zeros((100,100),dtype='i4')
  img=np.zeros((100,100),dtype='i4')
  fitsf = FITS('test.fits','rw')
  fitsf.write(img)
  img2=np.arange(3*3,dtype='i4').reshape(3,3)
  fitsf[0].write(img2, start=[2,2])
  fitsf.close()
```

Thanks,

Philippa

### PR DESCRIPTION
**Pull request by beckermr**
This PR address issue #284 where overwriting a rectangular subset of an image does not work as expected. I cannot tell if the change is breaking or not, given that the README is not clear and the doc string for the `start` keyword says:

```
start: integer or sequence
    Where to start writing data.  Can be an integer offset
    into the entire array, or a sequence determining where
    in N-dimensional space to start.
```

Given this and the fact that @esheldon seemed to think this was a bug in the original issue, I am working under the assumption this is an actual bug. 

In that case, this PR has a fix. 

~Note that unlike the original PR, I chose to iterate of contiguous slices of the array in python and make successive calls to the C layer instead of writing more C code.~ 

I wrote a C routine to access the `cfitsio` routine that writes a rectangular subset of an image. I also added tests.

closes #284

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -1838,7 +1838,7 @@ PyFITSObject_write_subset(struct PyFITSObject* self, PyObject* args) {
     }
 
     if (fits_movabs_hdu(self->fits, hdunum, &hdutype, &status)) {
-        set_ioerr_string_from_status(status);
+        set_ioerr_string_from_status(status, self);
         return NULL;
     }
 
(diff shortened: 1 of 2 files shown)


esheldon (approved):
Very nice, thank you

PR #464, opened 2025-10-22
fix: lossless compression of ints under PLIO

beckermr:
More compression bugs. See https://github.com/HEASARC/cfitsio/pull/99


esheldon (approved):
Thanks for this fix.

PR #475, opened 2025-11-26
feat: release the GIL

beckermr:
This PR is a rebase of #35 onto the current master so we can work on it.

closes #474 
closes #35

PR #476, opened 2025-11-26
feat: release the GIL very minimally

beckermr:
This PR is a retry of #35 and #475. 

~Instead of trying to release the GIL everywhere, which appears to break things, I am trying to only release for I/O critical operations.~

~To make further progress on better threaded performance, I think we'll need to separate out the cfitsio calls from the python C API calls so we can release the GIL for larger sections of code. My guess is that a bunch of time is being spent on initialization where the code goes back and forth.~

The approach here is to release the GIL for the biggest blocks of code as practically possible. There are some spots where this is hard because of the way python data structures are used to hold data as it is pulled from fitsio.

With this approach, I am seeing decent threaded performance once you do enough work in each thread. Threaded performance for reads is better than writes on my machine.


beckermr:
Hmmmm. I am not sure TBH since you'll need a compiler to build the package. :/

fitsio/tests/test_warnings.py
```diff
@@ -2,6 +2,7 @@
 import tempfile
 import warnings
 import numpy as np
+
```
beckermr:
```suggestion
```

beckermr:
The latest merge here went poorly. We'll want to break this PR up into smaller bits and proceed with single C functions one-by-one.

beckermr:
xref: https://github.com/esheldon/fitsio/pull/507

PR #481, opened 2025-12-09
feat: lock access to underlying cfitsio object

beckermr:
This PR adds locks to prevent concurrent usage of the cfitsio object.

See https://py-free-threading.github.io/porting/#raising-errors-under-shared-concurrent-use.

There are a lot of potential issues here, but the main ones are that 

- I have not structured the locks to ensure key operations are atomic.
- How the locks should be structured likely depends on the exact use case.

In other words, while I have ensured that no thread uses the same cfitsio file pointer at the same time, I have not ensured that a given thread can lock all of the python+C data structures long enough to do a fully consistent operation.

So for example, the lock is not held long enough for the thread to write image data via the C layer and also update the HDU list in the python layer.

So the locks as they are implemented now are formally correct, but don't really help the user  avoid important race conditions.

### DIFF
--- a/fitsio/fitsio_pywrap.c
+++ b/fitsio/fitsio_pywrap.c
@@ -37,13 +37,23 @@
 // max len of python error message
 #define PYFITS_ERRMSG_LEN 1024
 
+#if PY_MAJOR_VERSION >= 3 && PY_MINOR_VERSION >= 13
+#define LOCK(x) PyMutex_Lock(&(x->lock))
+#define UNLOCK(x) PyMutex_Unlock(&(x->lock))
+#else
+#define LOCK(x) (x)
+#define UNLOCK(x) (x)
+#endif
+
 struct PyFITSObject {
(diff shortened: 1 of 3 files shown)


fitsio/fitsio_pywrap.c
```diff
@@ -439,6 +440,34 @@ void append_string_to_list(PyObject* list, const char* str) {
...
 */
 
+static int check_py_fits_object_same_thread(struct PyFITSObject *self) {
+#ifdef PY_HAVE_THREAD_NATIVE_ID
+    unsigned long curr_thread_native_id;
+    curr_thread_native_id = PyThread_get_thread_native_id();
+
+    if (self->thread_native_id != curr_thread_native_id) {
+        PyErr_SetString(PyExc_RuntimeError,
+                        "Shared use of FITS object between objects detected! "
```
beckermr:
```suggestion
                        "Shared use of FITS object between threads detected! "
```

PR #491, opened 2026-05-02
test_compression_diskfile_kwargs: support both RICE_ONE and RICE_1

aurel32:
Starting with CFITSIO 4.6.4, commit 7d854ef29a4c ("Stop writing RICE_ONE as an alias for RICE_1."), the ZCMPTYPE header returns RICE_1 instead of RICE_ONE, breaking the test_compression_diskfile_kwargs test. Update it to accept both values to maintain compatibility across CFITSIO versions.


esheldon:
Thanks for the PR.

PR #493, opened 2026-05-08
Add second patch to imcompress.c limiting cache to one tile

### LINKED ISSUE #492: Limit memory usage of cached reads from compressed images
This cache can grow to be as big as the uncompressed data.

Let the file system do the cacheing.

this should be a simple patch

esheldon:
The old compressed tile cache could grow large.  This limits it to size one tile, addressing the main performance gain from the cache.

closes #492

### DIFF
--- a/patches/fitscore.c.patch
+++ b/patches/fitscore.c.patch
@@ -1,0 +1,62 @@
+--- fitscore.c	2025-09-22 11:49:36.000000000 -0400
++++ fitscore.c	2026-05-08 14:51:48.824886391 -0400
+@@ -4647,8 +4647,9 @@
+ 
+         /* free the tile-compressed image cache, if it exists */
+         if ((fptr->Fptr)->tilerow) {
+-           ntilebins = 
+-	    (((fptr->Fptr)->znaxis[0] - 1) / ((fptr->Fptr)->tilesize[0])) + 1;
++        /*    ntilebins =  */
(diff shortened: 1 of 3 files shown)


beckermr (changes requested):
I am not sure we should merge this fix. 

We should be carrying patches that fix critical bugs etc. We should not carry patches for new features. The intent is that the patch is eventually merged upstream.

Downstream users should probably close and reopen their fits file on a regular basis maybe instead?

esheldon:
I think of this as an implementation detail not a feature.

erykoff:
Right now this cfitsio "feature" is making it impossible to read DP2 healsparse maps in notebooks.  I was going to just change to astropy fits reading because of it.  Though I haven't checked what the overhead is of closing/opening.  So I'll compare the speed/memory of astropy vs close/open vs this patch.

beckermr:
OK. Even if you think this patch is an implementation detail, it needs to be combined with the existing patches etc.

esheldon:
This patch works as is.  Do you mean to do the other method instead, using a PR against cfitsio?

beckermr:
Yes. The code here is more or less a hack of the build system. I refactored the patching code to make it clearer and easier to understand.

esheldon:
OK, if we decide we want this I'll redo it.  Let's see what Eli finds

esheldon:
Superceded by #495

Issue #496, opened 2026-06-01
macos memory fault for writing PLIO_1 compressed image


esheldon:
In a CI job for a different package I used fitsio on macos to write a PLIO_1 compressed image and saw a memory fault

libmalloc bad-free in cfitsio::ffbinit

It is a flaky bug, some runs don't fail. It occurs using the binary installed from conda as well as locally built from source.

I did not observe a nano-malloc guard message

I don't have a mac to test, and I'll get diminishing returns trying to debug using a CI run.

Something like this might trigger it

```python
import os
import platform
import sys
import tempfile

import numpy as np

import fitsio


def env_banner():
    print(f"python:   {sys.version.split()[0]} ({platform.python_implementation()})")
    print(f"platform: {platform.platform()}")
    print(f"machine:  {platform.machine()}")
    print(f"fitsio:   {fitsio.__version__}")
    print(f"numpy:    {np.__version__}")
    try:
        print(f"cfitsio:  {fitsio.cfitsio_version()}")
    except Exception:
        pass
    print()


def write_plio1(tmpdir, i):
    """The crashing call, isolated."""
    fname = os.path.join(tmpdir, f"plio_{i:04d}.fits.fz")
    img = np.arange(16, dtype="i4").reshape(4, 4)
    with fitsio.FITS(fname, "rw") as f:
        f.write(img, compress="PLIO_1", tile_dims=(2, 2))


def write_rice1(tmpdir, i):
    """A 'control' compressed write that we know does NOT crash."""
    fname = os.path.join(tmpdir, f"rice_{i:04d}.fits.fz")
    img = np.arange(100, dtype="i4").reshape(10, 10)
    with fitsio.FITS(fname, "rw") as f:
        f.write(img, compress="RICE_1", tile_dims=(5, 5))


def run_alone(n):
    """N back-to-back PLIO_1 writes, fresh tempdir per call."""
    print(f"variant: PLIO_1 alone, {n} iterations")
    for i in range(n):
        with tempfile.TemporaryDirectory() as tmp:
            write_plio1(tmp, i)
        if (i + 1) % 10 == 0:
            print(f"  ok through {i + 1}")
    print("completed without abort")


def run_mixed(n):
    """
    Mirrors the test ordering that exposed the bug: many
    RICE_1 fixtures then a PLIO_1 write.  Loop the whole sequence.
    """
    print(f"variant: 25x RICE_1 then 1x PLIO_1, {n} outer iterations")
    for i in range(n):
[...]

PR #498, opened 2026-06-03
fix: need a minimum number of elements for PLIO compression

beckermr:
This PR adds a test from #496 to see if we can get the bug to trip.


esheldon:
Might be good to run with address sanitizing

CFLAGS="-fsanitize=address -g"

fitsio/fitsio_pywrap.c
```diff
@@ -1676,6 +1685,8 @@ static PyObject *PyFITSObject_create_image_hdu(struct PyFITSObject *self,
     }
 
     // this flushes all buffers
+    printf("about to flush file in Python C code\n");
+    fflush(stdout);
```
beckermr:
```suggestion
```

fitsio/fitsio_pywrap.c
```diff
@@ -20,6 +20,7 @@
     Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#include <stdio.h>
```
beckermr:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -91,7 +91,7 @@ jobs:
             python -c "import fitsio; assert not fitsio.cfitsio_has_bzip2_support()"
             python -c "import fitsio; assert not fitsio.cfitsio_has_curl_support()"
           else
-            pytest -vv ${PRP_FLAGS} fitsio
+            pytest -vvsx ${PRP_FLAGS} fitsio -k test_segfault_osx
```
beckermr:
```suggestion
            pytest -vv ${PRP_FLAGS} fitsio
```

esheldon:
In my tests this didn't happen on linux/x86_64 only the macos CI.  Did you get it to happen on both?

beckermr:
> In my tests this didn't happen on linux/x86_64 only the macos CI. Did you get it to happen on both?

No I did not. However, the address sanitizer pinpointed this bit of code as the buffer overrun and it appears to fix things.

beckermr:
The address sanitizer even pinpointed the error when running PLIO compression a single time.

beckermr:
OK @esheldon. This one is ready for review. I'd love to run the address sanitizer in CI, but I will leave that for another day. I went ahead and fixed a few other memory allocation bits that were not checked properly.

fitsio/fitsio_pywrap.c
```diff
@@ -2393,6 +2402,14 @@ static PyObject *PyFITSObject_write_columns(struct PyFITSObject *self,
     array_ptrs = calloc(ncols, sizeof(void *));
     nperrow = calloc(ncols, sizeof(LONGLONG));
     fits_dtypes = calloc(ncols, sizeof(int));
+    if ((is_string == NULL) || (colnums == NULL) || (array_ptrs == NULL) ||
+        (nperrow == NULL) || (fits_dtypes == NULL)) {
+        py_status = 1;
+        PyErr_SetString(
+            PyExc_MemoryError,
+            "Could not allocate metdata arrays for writing columns!");
```
beckermr:
```suggestion
            "Could not allocate metadata arrays for writing columns!");
```

esheldon:
I think the macos nano allocator is possibly catching this while glibc is letting it slide.

esheldon (approved):
Nice catches

PR #505, opened 2026-06-28
fix: add lock for python 3.13 or newer and update changelog for 1.4.0

beckermr:
See the PR title.


.github/workflows/tests-external-cfitsio.yml
```diff
@@ -10,7 +10,7 @@ env:
   PY_COLORS: "1"
   # These compiler flags force the tests to fail if arrays are
   # accessed at the C level from an unaligned location.
```
esheldon, line 12: update comment for ffp contract, and in other workflow

setup.py
```diff
@@ -371,7 +371,7 @@ def copy_update(dir1, dir2):
             '--disable-shared',
             '--enable-reentrant',
         ]
-        our_cflags = "-fPIC -fvisibility=hidden"
+        our_cflags = "-fPIC -fvisibility=hidden -ffp-contract=off"
```
esheldon: Add a comment

esheldon (changes requested):
This looks good except for needing some documentation.

I do think the FMA change should be in a separate PR.  If we ever need to revert something it would make it a lot easier.  But only if you have the energy for it.

beckermr:
done!

beckermr:
see https://github.com/esheldon/fitsio/pull/509

beckermr:
@esheldon ready for another look!

esheldon (approved):

PR #509, opened 2026-07-01
fix: do not use FMA instructions to ensure lossy floating-point compression is reproducible

beckermr:
On ARM systems, the use of FMA instructions causes lossy floating-point compression to differ from x86 systems. We turn these instructions off to ensure reproducibility.


beckermr:

Issue #513, opened 2026-07-04
Regression in 1.4.0: reading quantized/lossless-fallback float tiles flushes subnormal pixel values to 0.0


astrofrog:
Between 1.3.0 and 1.4.0 the value returned when reading a floating-point pixel whose stored bit pattern is a subnormal (denormal) changed from the actual subnormal value to `0.0`. This happens for float tiles that were stored via the lossless GZIP fallback path (i.e. tiles where quantization was skipped, so `ZSCALE = ZZERO = 0` and the data lives in `GZIP_COMPRESSED_DATA`).

The on-disk file is byte-for-byte identical between the two versions, so the change is entirely in the reader.

```python
import numpy as np, fitsio, os

data = np.zeros((5, 1), dtype='>f4')   # a single 5x1 tile, big-endian float32
data[:, 0] = [134.97459, 248.02034, 183.40105, 57.59670, 216.31425]

fn = "mini.fits"
if os.path.exists(fn):
    os.remove(fn)
with fitsio.FITS(fn, "rw") as f:
    f.write(data, compress="GZIP_1", tile_dims=(5, 1), qlevel=5, qmethod=-1)

back = fitsio.FITS(fn)[1].read()
print("fitsio  :", fitsio.__version__)
print("readback:", back.ravel())
```

Output with **1.4.0**:

```
fitsio  : 1.4.0
readback: [  0.       248.02034 183.40105  57.5967  216.31425]
```

Output with **1.3.0**:

```
fitsio  : 1.3.0
readback: [4.6095713e-41 248.02034 183.40105 57.5967 216.31425]
```

The last four values round-trip exactly in both versions; only the first element differs.

We came across this in the astropy test suite since we compare our results to fitsio.

beckermr:
This change was done to match a rejected patch from the upstream cfitsio maintianers.

beckermr:
If you'd like to advocate for a fix upstream, feel free to comment here: 

https://github.com/heasarc/cfitsio/pull/112

We try to not carry patches that change upstream behavior in fundamental ways. So we removed our patch once upstream rejected the PR. :/

astrofrog:
So just to check, do you mean this is intended behavior and consistent with cfitsio?

beckermr:
Yes it is. See the discussion on the issue I linked above.

beckermr:
See https://github.com/esheldon/fitsio/pull/514

astrofrog:
[...]

PR #514, opened 2026-07-05
fix: put back patches to prevent underflows from being cast to zero

beckermr:
This PR demonstrates that restoring the compression patches solves issue #513.


beckermr:
OK @esheldon This one is ready to go too. I think we should carry this patch even if upstream rejected it. The breakage with other FITS readers that do sane things is worse than carrying the patch.

esheldon (approved):

PR #516, opened 2026-07-05
fix: check memory allocation in C layer; drop support for python 2

### LINKED ISSUE #508: unchecked memory allocation via Python APIs
In the C layer there are a lot of calls like this:

```c
tobj = PyFloat_FromDouble(value);
PyDict_SetItemString(dict, key, tobj);
```

or like this

```c
dict = PyDict_New();
```

As far as I know, the values `tobj` and `dict` need to be checked if they are `NULL` and if so, an error needs to be raised.

cc @esheldon

beckermr:
As it says on the tin...

closes #508


beckermr:

esheldon (approved):

PR #517, opened 2026-07-06
fix: ensure hdu list and hdu map are in sync

beckermr:
This is a rare bug that occurred randomly during one of the recent CI runs. I do not know exactly how it happened, but the symptom was that `hdu_map` was missing while `hdu_list` was present. This PR ensures that if either is missing, which should never happen (but apparently did...), both are rebuilt.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,7 @@ Changes
 
 Bug Fixes
 
-- Fixed (rare) bug where `hdu_list` and `hdu_map` can go out of sync.
+- Fixed (rare) bug where `FITS.hdu_list` and `FITS.hdu_map` can go out of sync.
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -8,6 +8,10 @@ Changes
 - Reformatted `CHANGES.md` for clarity and added linting
   to ensure consistent formatting in the future.
 
+Bug Fixes
+
+- Fixed (rare) bug where `hdu_list` and `hdu_map` can go out of sync.
```
```suggestion
- Fixed (rare) bug where `FITS.hdu_list` and `FITS.hdu_map` can go out of sync.
```

beckermr:

esheldon (approved):

PR #518, opened 2026-07-06
fix: ensure we use a global lock on cfitsio calls when library is not reentrant

beckermr:
For non-reentrant builds, only one thread can call `cfitsio` functions at a time. The GIL has (I think) enforced this as a side effect since we did not release it. Now that we do release it in Python 3.13 or above, we need the extra guard. I also added a function to return whether the underlying cfitsio library is reentrant at the Python level for testing, debugging, etc.


beckermr:
@esheldon another quick one

PR #524, opened 2026-07-10
fix: lock updates to python object attributes

beckermr:
In free-threaded builds, I've seen what appear to be race conditions when `FITS` objects are shared across threads. See this test run: https://github.com/esheldon/fitsio/actions/runs/28881711342/job/85671386326?pr=516

This PR adds an RLock to prevent them. I am not sure if we should merge this. Instead, it might be better to simply not support this use case and remove the test that is failing.

### DIFF
--- a/fitsio/hdu/table.py
+++ b/fitsio/hdu/table.py
@@ -1000,7 +1000,6 @@ class TableHDU(HDUBase):
             If True, trim trailing spaces from strings. Will over-ride the
             trim_strings= keyword from constructor.
         """
-
         if keys:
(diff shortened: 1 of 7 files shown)


fitsio/hdu/table.py
```diff
@@ -983,6 +1000,7 @@ def read_rows(
             If True, trim trailing spaces from strings. Will over-ride the
             trim_strings= keyword from constructor.
         """
+
```
beckermr:
```suggestion
```

PR #525, opened 2026-07-10
fix: get threading support correct

beckermr:
It turns out that the lock in the C layer is not sufficient due to stored state in the Python wrappers. We can either add locks in the Python layer (see #524) or we can punt to the user. This PR punts to the user and removes the C layer locks for reentrant builds. Non-reentrant builds still need a global lock in the C layer.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ Changes
   underlying cfitsio library was compiled with reentrant support.
 - Added back patch of `cfitsio` to prevent subnormal floats from being
   cast to zero for bundled builds.
-- Updated docs to clarify threaded usage of `FITS` files.
+- Updated docs to clarify threaded usage of `FITS` objects.
 - Removed extra lock in C layer. Users must employ their own locks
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -19,6 +22,7 @@ Bug Fixes
 - Marked tests for reproducible lossy float compression as xfail for external
   cfitsio libraries.
 - Skip tests for large compressed images exceeding `2**32` bytes on 32-bit platforms.
+- Removed incorrect test of threaded usage.
```
```suggestion
- Fixed incorrect test of threaded usage.
```

CHANGES.md
```diff
@@ -11,6 +11,9 @@ Changes
   underlying cfitsio library was compiled with reentrant support.
 - Added back patch of `cfitsio` to prevent subnormal floats from being
   cast to zero for bundled builds.
+- Updated docs to clarify threaded usage of `FITS` files.
```
```suggestion
- Updated docs to clarify threaded usage of `FITS` objects.
```

PR #529, opened 2026-08-14
fix: use proper buffer size for PLIO

beckermr:
This PR adds a patch from https://github.com/heasarc/cfitsio/pull/174 to correct the PLIO buffer overrun. I am only adding part of the patch since it is easier to test this way and the rest of the patch did not apply correctly.


CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # Changes
 
+## 1.4.3
+
+Bug Fixes
+
+- Fixed further PLIO compression bugs using patch from https://github.com/heasarc/cfitsio/pull/174.
```
```suggestion
- Fixed further PLIO compression bugs using patch from [PR #174](https://github.com/heasarc/cfitsio/pull/174).
```

PR #531, opened 2026-09-17
fix: changes for big endian machines

bhatprachi:
The test `test_image_compression_gzip_subnormal_cast_to_zero` previously hardcoded `dtype='>f4'` (big-endian), which is already native on big-endian machines (s390x). The test relies on the built-in byte-swap of non-native data before writing, reinterpreting the bit pattern of the first value as a NaN in native byte order. That NaN is then written as zero in the compressed file and reads back as `0`. On s390x, the data was already native so no swap occurred, the bit pattern was never reinterpreted.

The fix detects native byte order at runtime and always creates the array in the opposite endianness, ensuring the byte-swap and the NaN reinterpretation occur on every platform.


esheldon:
thanks for the PR, looks good to me

